PR #28, opened 2016-01-29
Upgrade to core v6

Scooletz:
Migrating the core to v6.


Scooletz:
7 tests to go to full green + Address based tests.

SzymonPobiega:
Could not comment on actual files because there is too many diffs, but `AzureQueueNamingConvention` is not used.

SeanFeldman:
@SzymonPobiega `AzureQueueNamingConvention` is used here: https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/core-v6-migration/src/Transport/Config/AzureStorageQueueTransport.cs#L63

Scooletz:
All tests green. The only thing left is a possible change to queue creator. I need more information though, how it should be handled.
Ping @Particular/azure-maintainers if we can merge the current state to the develop, it'll be an enabler for all the other Core v6 related issues.

PR #33, opened 2016-02-08
Only receiving queries should be created when asked for creation.

Scooletz:
Only receiving queries should be created when asked for creation, sending queues will be created by receivers.

### DIFF
--- a/src/Transport/AzureMessageQueueCreator.cs
+++ b/src/Transport/AzureMessageQueueCreator.cs
@@ -28,10 +28,7 @@
                 await CreateQueue(address).ConfigureAwait(false);
             }
 
-            foreach (var address in queueBindings.SendingAddresses)
-            {
-                await CreateQueue(address).ConfigureAwait(false);
-            }
+            // the addresses that messages are sent to, should be created on its own, or by the receivers
         }
 
         private async Task CreateQueue(string address)


Scooletz:
Could you review & merge if ok? @SeanFeldman @yvesgoeleven

SzymonPobiega:
Why do you want to implement it like this? This is how MSMQ does it: https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/Msmq/QueueCreator.cs

TL;DR the sending queues are not all the queues an endpoint sends messages to. These only include queues like `audit` and `error`.

Scooletz:
@SzymonPobiega If you find it improper, probably I'm implementing this that way, because of having a wrong assumption about what receiving/sending queues are. Could we a quick sync tomorrow to discuss my assumptions? Before that I've been creating all the queries, in other words - queries from both sets. In a discussion with @yvesgoeleven & @SeanFeldman, it was mentioned that ASQ should create only queries that it receives from. This thought was the foundation of the change.

SeanFeldman:
For general review, please ping Particular/azure-maintainers.

SzymonPobiega:
I am fine as long as you guys are fine with the behavior. Just keep in mind that with that change the `error` and `audit` queues will no longer be created when running the installers. IMO this is a good change but is not something MSMQ users were accustomed to.

yvesgoeleven:
I see, I thought all destination queues were in sendingaddresses. If only audit & error are in there (and it stays that way), then we can create that set as well (as it was initially the case)

SzymonPobiega:
@yvesgoeleven there was never an intention to add sending addresses. In fact my original idea did not include the sending queues at all but we found out that this is necessary to preserve the `F5` behavior. 

That said, it might change if we decide to ship the file transport in V6 (not sure what was the final call). In such case we are free to remove the `F5` code all places but the development transport so the sending queues will go away. They are broken anyway -- on brokers there is a race condition and if more than one endpoint run installers in the same time exceptions might get thrown.

yvesgoeleven:
@SzymonPobiega than it's no problem, all I want to prevent is rogue destination queues

FYI:
- There is no install time in azure, we create everything at runtime
- Concurrency is a very normal thing in azure as every endpoint exists at least twice, so we're set up to handle that case

PR #34, opened 2016-02-09
Added coderules

distantcam:
Part of https://github.com/Particular/PlatformDevelopment/issues/610


Scooletz:
@distantcam I've removed the nonexistent `app.config` and amended your commit. All green now. Thank you for pushing this :)

Issue #25, opened 2016-01-25
FIPS compliance: Make MD5 algorithm replaceable with SHA1


yvesgoeleven:
Because MD5 algorithm, used for renaming queues when entity name is to large, is not FIPS compliant, we should make it replaceable with SHA1. Similar to https://github.com/Particular/NServiceBus.AzureServiceBus/pull/79

Scooletz:
@yvesgoeleven What about backward compatibility? How would you approach that? Use both and check if the old exists? Or pull from both?

yvesgoeleven:
This should be configurable, MD5 should probably stay the default? /cc @SeanFeldman

Scooletz:
Oh, if we can keep both algorithms and provide a choice, in my opinion we should use SHA1 and provide a configuration method `UseLegacyMD5QueueNameShortener` or similar.

SeanFeldman:
Can I suggest to avoid the word 'legacy'? Anything that is done at point in time `t` can become a legacy sooner or later.

As for the default - if an existing customer has queues in production shortened using MD5 and we'll use SHA1 based strategy by default in the new endpoints, MD5-based queues won't be found. It's fine as long as we we can intelligently let customers know what to do and how to fix it.

Scooletz:
Yes, I can remove that. I wanted explicitly connect with that `t` in the past, where we were using MD5. I can name it without legacy ofc, but this doesn't show why one should use MD5 or default SHA1 hashing (in the newest version).

yvesgoeleven:
honestly, nobody outside US government related projects will care about this, MD5 doesn't work when a certain FIPS related registry setting is set... not the case for 99% of our customers

Scooletz:
Will make `SHA1` opt in then, leaving `MD5` default.

PR #41, opened 2016-02-15
Make MD5 algorithm replaceable with SHA1

Scooletz:
Solves #25


Scooletz:
Ping @Particular/azure-maintainers Ready for a review & merge

src/Transport/Shortener.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        public static string Sha1(string test)
+        {
+            using (var provider = new SHA1CryptoServiceProvider())
+            {
+                var inputBytes = Encoding.Default.GetBytes(test);
+                var hashBytes = provider.ComputeHash(inputBytes);
+
+                return ToChars(hashBytes);
```
SeanFeldman: If you'd like to shorten, this would work: `return string.Join("", hashBytes.Select(x => x.ToString("x2")));`

Scooletz: This would make so many allocs. Can we keep it the way it is?

yvesgoeleven: Sure

Issue #27, opened 2016-01-29
Map namespace name to connectionstring


yvesgoeleven:
Connects to #22

In order to not leak security sensitive information in connectionstrings out, we need a map of logical level namespace name to physical level namespace connectionstring, so that we only need to pass the logical name around

- [ ] map logical namespace names to physical based on a partitioning strategy
- [ ] ensure only logical namespaces are used for routing and logging

PR #42, opened 2016-02-16
Mapping between namespaces and connection strings

Scooletz:
Connects to #27


Scooletz:
I have a few doubts @Particular/azure-maintainers 

1. Should this PR already cover extraction of a component translating from `destination` to `queue` + `namespace`
1. When dispatching, there's that validation step covering `Headers.ReplyToAddress`. I think that this might need covering in this PR as well, after all it's the endpoint which want to be sent a reply. Should this cover detecting if the message is sent to another namespace? Is that the place where we should use `connectionString` as we may run against node running core v5?

I hope you see where my questions are leading to. Anyway, looking for some feedback about the approach and the direction of this PR.

yvesgoeleven:
1. Yes, I think that logic warrants extraction into a dedicated component
2. Yes, we need to stay backward compat with V5, so we should have a configurable way to still send the connectionstring to those endpoints (but not throw away the other benefits like not logging the connectionstring)
3. Somewhere in the seam (ToTransportAddress iirc(?)) we're passing the destination part into the core, this should always be the mapped name so that we don't leak the credentials by accident in f.e. core logging.

Scooletz:
This clarifies a lot of my doubts. I'll leave this existing commit and work on top of it to address the points above.

src/AcceptanceTests/Sending/When_dispatching_to_other_account.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        public class Configuration : IConfigureTestExecution
+        {
+            public Task Configure(BusConfiguration configuration, IDictionary<string, string> settings)
+            {
+                configuration.UseTransport<AzureStorageQueueTransport>()
+                    .Addressing()
+                    .NamespacePartitioning()
+                    .AddNamespace(MainNamespaceName, Transports.Default.Settings["Transport.ConnectionString"]);
```
SeanFeldman: This is interesting. In native ASQ terminology there's an account. Should we keep this abstracted (to have it similar to ASB transport), or should we adhere to the native nomenclature? /cc @yvesgoeleven

src/Transport/AzureMessageQueueCreator.cs
```diff
@@ -53,6 +53,7 @@ public AzureMessageQueueCreator(CloudQueueClient client, QueueAddressGenerator a
                     if (info.HttpStatusMessage == "QueueAlreadyExists")
                     {
                         await TaskEx.CompletedTask.ConfigureAwait(false);
+                        return;
```
SeanFeldman: `return TaskEx.CompletedTask` ?

src/AcceptanceTests/Sending/When_dispatching_to_other_account.cs
boblangley: +1 for native nomenclature

src/Transport/AzureMessageQueueCreator.cs
Scooletz: True.

src/AcceptanceTests/Sending/When_dispatching_to_other_account.cs
Scooletz: That's my doubt as well. On one hand it eases the migration, on the other we loose 'native nomenclature'. I think that changing this slightly this to the following could be the best. It maintains the 'partitions' as a logical term and matches them with accounts.
```
 configuration.UseTransport<AzureStorageQueueTransport>()
	.Addressing()
	.NamespacePartitioning()
	.AddAccount(MainAccountName, Transports.Default.Settings["Transport.ConnectionString"]);	
```
Thoughts?

SeanFeldman: > .NamespacePartitioning()

We don't have namespaces in ASQ to partition :(

Scooletz: Then `Partitioning` :)

Scooletz:
I've introduced `QueueAtAccount` which is responsible for parsing and maintaining this information. The mapping between a name and a connection string is made by `AzureStorageAddressingSettings`. For now, this is backward compatible in a following way: if a user sends a message with `ReplyTo` header with a logical name, it will be translated to the account connection string as it was before.
Do we want to have this translation configurable? If yes, should it be opt-in as we did it for SHA1? Should user explicitly state
```
configuration.UseTransport<AzureStorageQueueTransport>()
    .Addressing()
    .Partitioning() 
    .UseOnlyLogicalNamesForAccounts() //<- sth like this?
    .AddStorageAccount("my", customConnectionString);
```
That could throw if a raw connection string is used.
I think that I need another pair of @Particular/azure-maintainers to take a look at this.

src/Transport/Dispatcher.cs
```diff
@@ -47,18 +44,21 @@ internal class Dispatcher : IDispatchMessages
...
-            var sendQueue = sendClient.GetQueueReference(addressGenerator.GetQueueName(queueName));
+            var q = addressGenerator.GetQueueName(queue.QueueName);
+            var sendQueue = sendClient.GetQueueReference(q);
 
             if (!Exists(sendQueue))
             {
                 throw new QueueNotFoundException
                 {
-                    Queue = queueName
+                    Queue = destination
```
SeanFeldman: Question: if `destination` can be in the format of `queue@destination`, should exception contain the `@destination` part? Assuming the exception will bubble up to the core and core will log it.

src/Transport/Dispatcher.cs
```diff
@@ -94,17 +94,19 @@ internal class Dispatcher : IDispatchMessages
...
 
-        // TODO: consider providing a more advanced mapping, providing ability to host queues in different storage accounts, without explicit sending the message to another one
-        // This could improve throughput enabling to a separate account for high-volume queueus. Additionally, one could come up with idea sharding the queue across different accounts.
-        // This could be done with a modified message pump pulling from different queues.
-        private string GiveMeConnectionStringForTheQueue(string queueName)
+        private string GetConnectionString(QueueAtAccount destination)
         {
-            return defaultConnectionString;
+            string connectionString;
+            if (addressing.TryMapAccount(destination.StorageAccount, out connectionString) == false)
```
SeanFeldman: If replying to a v7 (transport) endpoint, we should send back `queue@connStringName` rather than `queue@connString`. Looks like this won't happen if one forgets to map `connStringName` to a specific `connString`. 
So for new endpoints it sounds like we shouldn't fall back to to use `connString`. Saying that, for old endpoints we have to follow this logic. Should mapping be mandatory for all the accounts (`connString`s), allowing for endpoints to opt out of having `connStringName`s?

```c#
.AddStorageAccount("new-endpoints-namespace-1", "connstring1")
.AddStorageAccount("new-endpoints-namespace-2", "connstring2")
.AddStorageAccountWithoutName("connstring2")
```

By this we're mandating to register all the storage accounts involved, yet we can control when to throw an exception warning a person and when not. That exception could also guide, by saying "either this is an old endpoint that only goes by `connString` or a new endpoint that should be registered using the following syntax...".

Scooletz: How one can tell, that endpoint that will receive the message is in a given version and will understand the new mapping? If there's any way of checking this, I'm not familiar with it.

If fallow a mandatory path for setting all the accounts, then mapping becomes much cleaner. I'm :+1: for this.

src/Transport/Dispatcher.cs
Scooletz: Assuming, that we're mapping all the accounts, I can easily check that the destination is passed only when an account is mapped.

src/Transport/Dispatcher.cs
SeanFeldman: @Scooletz yes, that would make registration mandatory for the new endpoints.

src/Transport/Dispatcher.cs
SeanFeldman: :+1:

Scooletz:
I've applied the mapping between logical name and the connection string. Now the `MessagePump` migrates v6 messages to v7 so whole pipeline uses v7. When sending a message, a flag `UseLogicalQueueAddresses` is checked if a new address format can be used. If any of the endpoints running, still use v6 don't set this flag. When all are migrated, set it to enable the new address format.

This finishes all requirements of #27

Ping @Particular/azure-maintainers for the review and possibly, merge to develop.

src/AcceptanceTests/app.config
```diff
@@ -3,19 +3,19 @@
   <runtime>
     <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
       <dependentAssembly>
-        <assemblyIdentity name="NServiceBus.Core" publicKeyToken="9fc386479f8a226c" culture="neutral" />
+        <assemblyIdentity connectionStringValue="NServiceBus.Core" publicKeyToken="9fc386479f8a226c" culture="neutral" />
```
dvdstelt: I highly doubt this should be connectionstring?! :)

src/Tests/app.config
```diff
@@ -3,15 +3,15 @@
   <runtime>
     <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
       <dependentAssembly>
-        <assemblyIdentity name="Microsoft.Data.Edm" publicKeyToken="31bf3856ad364e35" culture="neutral" />
+        <assemblyIdentity connectionStringValue="Microsoft.Data.Edm" publicKeyToken="31bf3856ad364e35" culture="neutral" />
```
dvdstelt: Same, likely not a connectionstring

src/Transport/ConnectionString.cs
```diff
@@ -0,0 +1,59 @@
+﻿namespace NServiceBus.Azure.Transports.WindowsAzureStorageQueues
+{
+    using System;
+
+    public sealed class ConnectionString : IEquatable<ConnectionString>
```
yvesgoeleven: should this be public?

src/Transport/Config/QueueIndividualizer.cs
```diff
@@ -1,38 +1,41 @@
 namespace NServiceBus.Azure.Transports.WindowsAzureStorageQueues
 {
+    using System;
     using System.Globalization;
+    using NServiceBus.Azure.Transports.WindowsAzureStorageQueues.Config;
     using Support;
 
-    class QueueIndividualizer
+    public class QueueIndividualizer
```
yvesgoeleven: Should this be public?

src/Transport/Config/QueueAddress.cs
```diff
@@ -0,0 +1,92 @@
+﻿namespace NServiceBus.Azure.Transports.WindowsAzureStorageQueues.Config
+{
+    using System;
+
+    public sealed class QueueAddress : IEquatable<QueueAddress>
```
yvesgoeleven: should this be public?

src/Tests/app.config
yvesgoeleven: It amazes me that the tests even run because of this, how can that be?

src/Transport/Config/Extensions/AzureStorageAddressingSettings.cs
```diff
@@ -0,0 +1,129 @@
...
+
+        internal void SetDefaultAccountConnectionString(string connectionString)
+        {
+            Add(QueueAddress.DefaultStorageAccountName, connectionString, false);
+        }
+
+        /// <summary>
+        ///     Transforms all the <see cref="QueueAddress" /> in <see cref="headersToApplyNameMapping" /> to use logical names.
+        /// </summary>
+        internal void ApplyMappingToLogicalName(HeadersCollection headers)
```
melkio: Is this a "settings" concern? Isn't it better to move this logic away from settings?

src/Tests/app.config
SeanFeldman: Latest versions of assemblies are in place, so no redirect is used? Therefore we could get rid of these.

src/Transport/Config/QueueAddress.cs
SeanFeldman: In not planned to be used to build additional topologies (can't recall that requirement for ASQ), then shouldn't.

src/Tests/app.config
Scooletz: Will do.

src/Transport/ConnectionString.cs
Scooletz: That's an overall question. Should I aim at keeping the transport as `internal` as possible?

src/Transport/Config/Extensions/AzureStorageAddressingSettings.cs
Scooletz: I can't see the owner of this now. I can move it to an `internal` extension method.

Scooletz:
It looks like many of the comments above are touching the `public` implementations. Can we postpone it for now as #43 is planned for v7 of this transport as well?

Scooletz:
I've covered the `app.config` files + the method setting the default. As I mentioned above, I'd like to keep the visibility changes for #43.
Ping @Particular/azure-maintainers for the review.

src/Transport/Config/QueueAddress.cs
```diff
@@ -0,0 +1,92 @@
...
+            QueueAddress q;
+            if (TryParse(value, out q) == false)
+            {
+                throw new ArgumentException("Value cannot be parsed", nameof(value));
+            }
+
+            return q;
+        }
+
+        public static bool TryParse(string inputQueue, out QueueAddress queue)
```
melkio: Why `TryParse` if always returns `true`?

src/Transport/Config/Extensions/AzureStorageAddressingSettings.cs
melkio: Design of setting classes in ASB is completely different: settings are only to store "settings". Specific behaviors are moved away

src/Transport/Config/Extensions/AzureStorageAddressingSettings.cs
```diff
@@ -0,0 +1,124 @@
...
+                }
+            }
+        }
+
+        private bool TryMap(ConnectionString connectionString, out string name)
+        {
+            return _connectionString2name.TryGetValue(connectionString, out name);
+        }
+
+        internal void Add(string name, string connectionString, bool add = true)
```
melkio: a method with a bool flag "are" two methods :)

SeanFeldman: Optional `bool` :wink: 
It's an internal method, so can definitely see how this could be used for a test needs or something similar.

src/Transport/ConnectionString.cs
SeanFeldman: :+1:

src/Transport/Config/Extensions/AzureStorageAddressingSettings.cs
Scooletz: This method is similar to the `Dictionary<TKey, TValue>.Insert` which accepts `add` parameter to check if it should throw on an existing item. If there was a big `if` in the middle, I'd agree, but in this case that's a precondition check flag.
@SeanFeldman IMHO it's ok to use optional parameters for a nonpublic API. I'm aware of the mess it can bring to the table when used in a public surface :P

src/Transport/Config/QueueAddress.cs
Scooletz: It's a valid question, but I've found no case where the value could be not accepted. I can see now that it should verify the `queueName` as it may throw on empty. Will add this check and return `false`. I see no other scenario.

Scooletz:
ping @Particular/azure-maintainers for the review. I'm aware of broken tests and will cover that, but I want to close it from the API perspective (timeouts in tests is a separate thing, but will resolved as a part of this pr)

yvesgoeleven:
1 thing I would change to the API: rename UseOnlyLogicalNamesForAccounts to UseAccountNamesInsteadOfConnectionStrings

melkio:
@yvesgoeleven : should be the ASB and ASQ config API the same? In ASB we haven't "Account" concept. I called it `UseLogicalNamespaceName`

yvesgoeleven:
@melkio consistency would be nice, but use domain specific term. UseNamespaceNamesInsteadOfConnectionStrings

Scooletz:
Should I squash all of these commits?

The tests are green. The reason behind time outs was abusive cleaning queues running through all of them. After applying the Msmq way of cleaning queues (it uses `QueueBindings` & `QueueAddressGenerator` now to select only these used in a specific test), the tests run much faster... and are green :)

yvesgoeleven:
I'm no big fan of squashing, I rather see what happened

Issue #24, opened 2016-01-25
Code only configuration API


yvesgoeleven:
Connects to #22

To be consistent with core v6 & the rest of the transports & persistences, we should switch to a code only configuration model on top of SettingsHolder.

Todo: should we provide an overlay for backward compat of xml based config model as well?

yvesgoeleven:
This item replaces https://github.com/Particular/NServiceBus.AzureStorageQueues/issues/15

Scooletz:
Should the old config section be removed or marked with `ObsoleteEx` for now? I cannot find any rule for that beside https://github.com/Particular/Vision/issues/22

SzymonPobiega:
Well, we need to somehow tell the users that the values there are no longer used in any way...

yvesgoeleven:
I noticed core is also going to remove all config sections, what's the plan there to tell users about it?

SeanFeldman:
@SzymonPobiega I wonder about the same https://github.com/Particular/NServiceBus.AzureStorageQueues/issues/24#issuecomment-180526821

dvdstelt:
The original item #15 said ConfigSection should take precedence over configuration done in code. That's not what I'm used to, from other Microsoft libraries. For example in WCF everything in config is overridden by code configuration.

Scooletz:
I agree with @dvdstelt. I'm used to **Code over configuration** rule as well.

dvdstelt:
And that's also what this says : https://github.com/Particular/Vision/issues/22#issuecomment-105154194

tmasternak:
@dvdstelt given that `Code over config` PR has been merged is there anything else left to do here? If yes let's move that to different issue and close this one. It's not a must-have for V6 release, right?

PR #48, opened 2016-03-02
Code over config

dvdstelt:
Connected to https://github.com/Particular/NServiceBus.AzureStorageQueues/issues/24

Removed configsection & appsetting "transport/nservicebus"
Everything is now code-only.


src/Transport/Config/AzureStorageTransportExtensions.cs
```diff
@@ -30,7 +46,7 @@ public static TransportExtensions<AzureStorageQueueTransport> PeekInterval(this
         /// <returns></returns>
         public static TransportExtensions<AzureStorageQueueTransport> MaximumWaitTimeWhenIdle(this TransportExtensions<AzureStorageQueueTransport> config, int value)
         {
-            config.GetSettings().Set(ReceiverMaximumWaitTimeWhenIdle, value);
+            config.GetSettings().Set(WellKnownConfigurationKeys.ReceiverMaximumWaitTimeWhenIdle, value);
```
SeanFeldman: The old code was not necessarily guarding against invalid parameter values, but should we?

src/Transport/Config/AzureStorageTransportExtensions.cs
```diff
@@ -65,7 +81,7 @@ public static TransportExtensions<AzureStorageQueueTransport> MessageInvisibleTi
         /// <returns></returns>
         public static TransportExtensions<AzureStorageQueueTransport> BatchSize(this TransportExtensions<AzureStorageQueueTransport> config, int value)
         {
-            config.GetSettings().Set(ReceiverBatchSize, value);
+            config.GetSettings().Set(WellKnownConfigurationKeys.ReceiverBatchSize, value);
```
SeanFeldman: Same as in https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/48/files#r54795389

src/Transport/Config/AzureStorageTransportExtensions.cs
```diff
@@ -96,46 +112,21 @@ public static TransportExtensions<AzureStorageQueueTransport> SerializeMessageWr
         /// </summary>
         public static TransportExtensions<AzureStorageQueueTransport> CreateSendingQueues(this TransportExtensions<AzureStorageQueueTransport> config)
         {
-            config.GetSettings().Set(TransportCreateSendingQueues, true);
+            config.GetSettings().Set(WellKnownConfigurationKeys.TransportCreateSendingQueues, true);
```
SeanFeldman: Should we let users specify via parameter if creation should take place or not? We can default to `true` if we want, but allow to pass in `false`.

src/Transport/Config/AzureStorageTransportExtensions.cs
```diff
@@ -96,46 +112,21 @@ public static TransportExtensions<AzureStorageQueueTransport> SerializeMessageWr
...
             return config;
         }
 
         public static TransportExtensions<AzureStorageQueueTransport> UseSha1ForShortening(this TransportExtensions<AzureStorageQueueTransport> config)
         {
-            config.GetSettings().Set(Sha1Shortener, true);
+            config.GetSettings().Set(WellKnownConfigurationKeys.Sha1Shortener, true);
```
SeanFeldman: Same as [here](https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/48/files#r54796221)

src/Transport/Config/AzureStorageTransportExtensions.cs
dvdstelt: Do you mean as an overload (or default value), providing both "usethis()" and "usethis(false)" options?

SeanFeldman: @dvdstelt now that I look at this, this method should be gone. Decision was made that destination queues should not be created by the sender. As such, this becomes obsolete. But for other methods like this, yes, your comment is correct.

dvdstelt: Even better, removing it. I'll make it so.

dvdstelt:
I'm updating to NSB6 1616 where some stuff was changed, Initialize method added, etc.

Because when updating TransPortCompat tests to latest NSB version, it was complaining about that Initialize method and more. And changes need to be done anyhow.

Issue #47, opened 2016-03-02
Bump up the core once again


Scooletz:
Connects to #22

We need to bump up the core once again as there have been changes introduced on the transport seam level.

dvdstelt:
I'm blocked picking up this : https://github.com/Particular/ProductionTests/issues/45
So I'll do this first.

andreasohlund:
@Particular/transport-maintainers since @Scooletz is busy with the saga persister can we hand this one over to some one else?

dvdstelt:
This is ready for merge

yvesgoeleven:
Merged

PR #50, opened 2016-03-04
Upgrade to latest V6 core

dvdstelt:
Connected to #47 

Upgraded to new TransportDefinition and EndpointConfiguration


dvdstelt:
/cc @Particular/azure-maintainers please review

There are quite a few commits in here that could be squashed, but not sure if that's possible if they're already on remote?

src/Transport/Config/DefaultConfigurationValues.cs
```diff
@@ -0,0 +1,33 @@
+﻿namespace NServiceBus.AzureStorageQueues
+{
+    using NServiceBus.Azure.Transports.WindowsAzureStorageQueues.Config;
+    using NServiceBus.Settings;
+
+    class DefaultConfigurationValues
+    {
+        // These are here only for the ConfigSection, can be removed in V7
```
SeanFeldman: old comment?

dvdstelt: You're right, will remove

Scooletz:
@dvdstelt You can squash and then 'push --force' it to the remote. As long as it's a separate PR branch, you can do anything. If you need any help with this, ping me.

SimonCropp:
the nuspec is still wrong
https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/UpdateToV6/packaging/nuget/nservicebus.azurestoragequeues.nuspec#L20

needs to include the unstable of v6 eg https://github.com/Particular/NServiceBus.Callbacks/blob/develop/packaging/nuget/nservicebus.callbacks.nuspec#L22

also note there are more tokens that can be taken advantage of eg `<copyright>$copyright$</copyright>`

PR #52, opened 2016-03-11
Cleaning queues moved from the clean up to the test configuration

Scooletz:
To address various problems when changing configurations and obtaining values from the previous run.


Scooletz:
Ping @Particular/azure-maintainers , especially @dvdstelt :-)

Issue #60, opened 2016-03-30
Queue purge causing ATTs to be flaky


SeanFeldman:
There's something about `When_timeout_dispatch_fails.Should_retry_and_move_to_error` that is flaky or the cleanup is not working properly. All tests/ATTs are passing locally (using Azure storage account). 

This issue has surfaced up when working on PR https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/56

Scooletz:
:+1: I've been hit by this on develop. I've seen that the failing was caused by either less or greater number of messages. In one case, after ending the test a message was left in the queue.

A failure example:
https://builds.particular.net/viewLog.html?buildId=155325&tab=buildResultsDiv&buildTypeId=NServiceBus_Transports_NServiceBusAzureStorageQueues_2RunAcceptenceTests

SeanFeldman:
For ASQ, is cleanup taking at the end of each test or prior to test execution?

dvdstelt:
Prior to

SeanFeldman:
After looking into the [`CleanQueuesUsedByTest` ](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/aadad168fc466edee19fe47420f477cf468e8d88/src/AcceptanceTests/ConfigureAzureStorageQueueTransport.cs#L48) execution with a debugger, I've noticed that [`GetTestRelatedQueueNames`](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/aadad168fc466edee19fe47420f477cf468e8d88/src/AcceptanceTests/ConfigureAzureStorageQueueTransport.cs#L65) resolved no queue names... This could be an issue for the cleanup.

SeanFeldman:
I have traced through and think the issue is that ATT endpoints are started _after_ cleanup is executed. Hence the empty `QueueBindings`. Wonder if this is a recent change in ATTs/core.

dvdstelt:
You merged it 20 days ago! :)
https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/52

SeanFeldman:
@dvdstelt what I meant is "if there _was_ a change to ATTs/core _recently_". I assume you guys had it working.

Scooletz:
[...]

PR #62, opened 2016-04-01
Clear all the queues before every test

Scooletz:
Connects to #60
solving it by clearing all the queues before every test

### DIFF
--- a/src/AcceptanceTests/ConfigureAzureStorageQueueTransport.cs
+++ b/src/AcceptanceTests/ConfigureAzureStorageQueueTransport.cs
@@ -1,14 +1,14 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.WindowsAzure.Storage;
+using Microsoft.WindowsAzure.Storage.Queue;
 using NServiceBus;
 using NServiceBus.AcceptanceTesting.Support;
 using NServiceBus.AcceptanceTests.ScenarioDescriptors;
 using NServiceBus.Azure.Transports.WindowsAzureStorageQueues;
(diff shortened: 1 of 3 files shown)


Scooletz:
Ready to be merged.
Ping @Particular/azure-maintainers @dvdstelt

src/AcceptanceTests/ConfigureAzureStorageQueueTransport.cs
```diff
@@ -45,28 +41,28 @@ public Task Cleanup()
...
         {
-            var queue = queues.GetQueueReference(queuesName);
-            if (await queue.ExistsAsync())
-            {
-                await queue.ClearAsync();
-            }
+            queue.ClearAsync().ContinueWith(t => countdown.Signal());
+        }
+
+        if (countdown.Wait(TimeSpan.FromMinutes(1)) == false)
```
dvdstelt: Nice solution!

SeanFeldman:
Purging all queues approach is going to be challenging as `IConfigureEndpointTestExecution` methods (`Configure` and `Cleanup`) are involved per each endpoint participating in acceptance test. That means we'll clean all the queues N times (N is the number of endpoints per ATT).

Don't know if NUnit can indicate when the whole test suite is starting or ending. If that would be possible, it would be an ideal place to hook into.

Scooletz:
@SeanFeldman The one way of doing is providing `ITestAction` http://nunit.org/index.php?p=actionAttributes&r=2.6.2 but this requires applying it on all the suites. The other approach is using the provided extensibility point & by writing a small plugin implementing http://nunit.org/index.php?p=testDecorators&r=2.6.2 run a piece of code before every suite. Time: up to 1 day.

PR #70, opened 2016-04-11
Cleanup

SimonCropp:
* remove redundant code
 * update obsolete and approval tests			
 * remove Microsoft.WindowsAzure.Configuration  …


SeanFeldman:
@SimonCropp  approved.txt should be committed?

SimonCropp:
@SeanFeldman nah i just reverted that change to the public API. we can handle that in another PR where review all public APIS

SeanFeldman:
That API might change - there's a way now to set a default for the transport. Plus, those are in a nested namespace which makes discovery almost impossible for users.

PR #75, opened 2016-04-15
Remove uneeded dispatch failed exception

### LINKED ISSUE #83: UnableToDispatchException.Namespace is never set
should it be?

andreasohlund:
fix #83


andreasohlund:
Ping @Scooletz

src/Transport/MessagePump.cs
```diff
@@ -195,6 +184,7 @@ public void Start(PushRuntimeSettings limitations)
                 catch (Exception ex)
                 {
                     await retrieved.Nack().ConfigureAwait(false);
+                    await circuitBreaker.Failure(ex).ConfigureAwait(false);
```
WojcikMike: @andreasohlund originally you had here: TODO: enable cb
So I did it. If there were any reasons why you put TODO instead of code, let me know

@Scooletz have a look at this line, as I am almost sure it is right (like 70%)

andreasohlund: :+1:

WojcikMike:
@Particular/azure-storage-queues-maintainers please review and merge

src/Transport/MessagePump.cs
Scooletz: I'd move `cb.Failure` before `Nack`. If `Nack` fails, then it throws an exception before notifying cb about the failure. The exception is captured by the task that later on is continued with https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/remove-dispatchfailed-exception/src/Transport/MessagePump.cs#L143 and marked as `Success`. Having that said, I'd place failure before the Nack.

src/Transport/MessagePump.cs
```diff
@@ -208,6 +197,7 @@ static int EstimateDegreeOfReceiveParallelism(int maxConcurrency)
...
                     Logger.Warn("Azure Storage Queue transport failed pushing a message through pipeline", ex);
```
Scooletz: I'd move this before `Nack` as well.

Issue #67, opened 2016-04-10
Performance review


SeanFeldman:
Connects to https://github.com/Particular/PlatformDevelopment/issues/723

The End2End tests show a drop in a performance. An ASQ Transport specialist should take a look into it and profile the transport. The outcome of this could be a list of fixes that need to be applied and/or documentation that needs to updated/created to point users how to improve performance.

Comparison was done between v6 and v7 using EndToEnd testing.
https://particularsoftware.slack.com/archives/tf-v6-perf-tests/p1460293006000468

Consider following hints when working on this:
- is `JsonSerializer` explicitly set for the ASQ transport. Before `core v6`, transports could augment the serialization choice and `ASQ` was changing it to `JSON`. Currently, it needs to be stated explicitly using ` .SerializeMessageWrapperWith<JsonSerializer>();`
- max number of concurrent http connections using either [config section](https://msdn.microsoft.com/en-us/library/fb6y0fyc.aspx) or [System.Net.ServicePointManager.DefaultConnectionLimit](http://msdn.microsoft.com/en-us/library/system.net.servicepointmanager.defaultconnectionlimit.aspx)
- batch size

`/cc @Particular/azure-maintainers

Scooletz:
This is just a remark. We still need the tests.

---
The biggest gain I can think of is `NACK`ing messages using `UpdateMessage` to descrease the `visibilityTimeout` when a failure occurs. In v6 message is invisible for the default invisibility time, currently, we're going to retry much faster.

andreasohlund:
Will this be needed for beta or rc?

WilliamBZA:
@Scooletz @SeanFeldman the code for the tests can be seen over [here](https://github.com/Particular/EndToEnd/pull/8). We'll update this issue once we get more results.

Scooletz:
@WilliamBZA Great!

tmasternak:
@Particular/tf-v6-launch-plat-dev I think that we need to flesh out the scope of this a bit better. That said I think this needs to go into `V6 Launch`

boblangley:
@tmasternak I agree. @andreasohlund asks my question. Perhaps part of finding the answer would be the questions: How bad is the performance drop? Could we go to beta as is?

Scooletz:
@tmasternak @boblangley Then I'd propose just to move to the beta preparation and deal with it.

timbussmann:
[...]

PR #77, opened 2016-04-15
MessagePump and Message Ack changes

danielmarbach:
Connects to #67

* MessagePump no longer uses Task.Run for async receive and no longer schedules a task for tokenSource.Dispose
* Ack of messages happens concurrently in the inner receive and only if there are slot on the semaphore available
* Unpacking the message envelop concurrently in the inner receive

### DIFF
--- a/src/Transport/MessagePump.cs
+++ b/src/Transport/MessagePump.cs
@@ -14,9 +14,6 @@
 
     class MessagePump : IPushMessages, IDisposable
     {
-        AzureStorageAddressingSettings addressing;
-
-        AzureMessageQueueReceiver messageReceiver;
         public MessagePump(AzureMessageQueueReceiver messageReceiver, AzureStorageAddressingSettings addressing)
         {
             this.messageReceiver = messageReceiver;
@@ -151,7 +148,7 @@
                     Logger.Error($"The dispach failed at sending a message to the following queue: '{ex.Queue}'", ex);
                     await circuitBreaker.Failure(ex).ConfigureAwait(false);
                 }
-                catch (OperationCanceledException)
+                catch (TaskCanceledException)
                 {
                     return;
                 }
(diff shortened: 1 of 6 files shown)


danielmarbach:
Will rebase as soon as https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/84 is in

src/Transport/MessageEnvelopUnpacker.cs
```diff
@@ -0,0 +1,52 @@
+namespace NServiceBus.Azure.Transports.WindowsAzureStorageQueues
+{
+    using System;
+    using System.IO;
+    using System.Runtime.Serialization;
+    using Microsoft.WindowsAzure.Storage.Queue;
+
+    class MessageEnvelopUnpacker
```
Scooletz: What's the reason for introducing another entity responsible for serialization/deserialization? Why don't we put this logic (if needed) to the `MessageWrapperSerializer` itself?

src/Transport/MessagePump.cs
```diff
@@ -140,71 +156,58 @@ public void Start(PushRuntimeSettings limitations)
...
-                var tokenSource = new CancellationTokenSource();
-                var receiveTask = Task.Run(async () =>
+        private async Task InnerReceive(MessageRetrieved retrieved)
+        {
+            using (var tokenSource = new CancellationTokenSource())
+            {
+                try
                 {
-                    try
+                    if (ackBeforeDispatch)
```
Scooletz: Moving closer to the real handling and doing it in a task itself - nice!

src/Transport/MessagePump.cs
```diff
@@ -140,71 +156,58 @@ public void Start(PushRuntimeSettings limitations)
...
-                        var message = retrieved.Wrapper;
+                        await retrieved.Ack().ConfigureAwait(false);
+                    }
 
-                        addressing.ApplyMappingToLogicalName(message.Headers);
+                    var message = retrieved.Unpack();
+                    addressing.ApplyMappingToLogicalName(message.Headers);
 
-                        var pushContext = new PushContext(message.Id, message.Headers, new MemoryStream(message.Body), new TransportTransaction(), tokenSource, new ContextBag());
+                    using (var memoryStream = new MemoryStream(message.Body))
```
Scooletz: From the technical point of view, it does not change anything, but I see, that we should be disposing things.

src/Transport/MessageEnvelopUnpacker.cs
danielmarbach: Previously that code was in the receiver. I just extracted it out from the receiver and put it into a named entity. Since this entity deals with CloudQueueMessages it feels off to put it into the serializer.

src/Transport/MessagePump.cs
danielmarbach: plus now it is concurrent :dancers:

src/Transport/MessageEnvelopUnpacker.cs
Scooletz: I see. 
One spelling error. It should be `MessageEnvelop**e**Unpacker`, shouldn't it? What about `MessageEnvelopeUnwrapper` - would it be better?

Scooletz:
So the parallelism will be a separate PR?

danielmarbach:
Yes parallelism will be done in a distinct PR

PR #78, opened 2016-04-16
SyncOMatic update

particularbot:


danielmarbach:
Before we merge this we probably need to cleanup the whole solution

danielmarbach:
Replaced by #84

PR #80, opened 2016-04-16
[SPIKE] Message pump and unpack in parallel

danielmarbach:
Connects to #67

Description will follow


danielmarbach:
Unpack in parallel not necessary. Will close this and update #79

PR #84, opened 2016-04-18
Code style cleanup

danielmarbach:
Connects to #67

Replaces https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/78


danielmarbach:
@Particular/azure-maintainers can you review and merge if green?

src/Transport/Config/DefaultConfigurationValues.cs
```diff
@@ -28,5 +20,13 @@ void ApplyDefaults(SettingsHolder settings)
             settings.SetDefault(WellKnownConfigurationKeys.PurgeOnStartup, DefaultPurgeOnStartup);
             settings.SetDefault(WellKnownConfigurationKeys.DefaultQueuePerInstance, DefaultQueuePerInstance);
         }
+
+        const int DefaultMessageInvisibleTime = 30000;
```
melkio: Why aren't these settings inlined in the `SetDefault` call? For me setup should be more readable

src/Transport/ConnectionString.cs
```diff
@@ -48,12 +46,14 @@ public override bool Equals(object obj)
...
-            return Equals((ConnectionString)obj);
+            return Equals((ConnectionString) obj);
         }
 
         public override int GetHashCode()
         {
             return Value.GetHashCode();
         }
+
+        public readonly string Value;
```
melkio: It's an internal field. Should it be `value` (lowercase)?

src/Transport/CreateQueueClients.cs
```diff
@@ -28,5 +26,7 @@ CloudQueueClient BuildClient(ConnectionString connectionString)
 
             return null;
         }
+
+        ConcurrentDictionary<ConnectionString, CloudQueueClient> destinationQueueClients = new ConcurrentDictionary<ConnectionString, CloudQueueClient>();
```
melkio: Since ASQ doesn't support multiaccounts (yet) why are we using a `ConcurrentDictionary`?

melkio:
@danielmarbach a few comments added here and there :)

danielmarbach:
@melkio thanks for the review. The cleanup did not change the code but just move around field and methods according to the style template. So how about we address your points in a dedicated issue? Can you open one?

melkio:
ok @daniel I agree with you on the the third comment. There are two others comments related to "style/cleanup". Are you suggesting to create an issue to rename `Value` field to `value`? Or to inline default values? 
Isn't it better to address them here? Let me know what you prefer...

danielmarbach:
@melkio since you are part of the maintainers I'd just do it as a PR and get it in.

Issue #87, opened 2016-04-18
Review API change and introduce Obsolete


WojcikMike:
Related to: https://github.com/Particular/PlatformDevelopment/issues/723

We should review the v6 API and mark with obsolete all public members that were removed. This will probably feed into Upgrade guide for ASQ: 
Particular/docs.particular.net#1275 
Particular/docs.particular.net#1423

WojcikMike:
Code and doco updated, closign this issue

PR #90, opened 2016-04-19
Marking as obsolete all changes in public API

WojcikMike:
Connects to #87

Related to doco PR: https://github.com/Particular/docs.particular.net/pull/1437


WojcikMike:
@Particular/azure-storage-queues-maintainers  I need some help on Obsolete messages, please have a look at obsolete.cs file as a lot of types were removed or made internal. I usually don't have good explanation on why, so I put a generic text saying:
"This class served only internal purposes without providing any extensibility point and as such was removed from teh public API, please refer to the documentation for more information."

I will put into documentation information about changing in our public API with regards about keeping public what only is needed for users

src/Tests/API/APIApprovals.ApproveAzureStorageQueueTransport.approved.txt
```diff
@@ -1,6 +1,41 @@
 ﻿namespace NServiceBus.Azure.Transports.WindowsAzureStorageQueues
 {
     
+    [System.ObsoleteAttribute("This class served only internal purposes without providing any extensibility poin" +
```
SeanFeldman: > removed from teh public API
type "teh" => "the"

src/Tests/API/APIApprovals.ApproveAzureStorageQueueTransport.approved.txt
```diff
@@ -1,6 +1,41 @@
...
+        "t and as such was removed from teh public API, please refer to the documentation" +
```
SeanFeldman: How about turning 

> , please refer to the documentation...

into  ". For more information, refer to the documentation."?

src/Transport/AzureMessageQueueCreator.cs
```diff
@@ -1,4 +1,4 @@
-﻿namespace NServiceBus.Azure.Transports.WindowsAzureStorageQueues
+﻿namespace NServiceBus.Azure.Transports.WindowsAzureStorageQueues.Internals
```
SeanFeldman: Have to admit I'm not a big fan of ".Internals" namespace.
Would rather see a calls marked with accessibility modifier, which seems to be a no-go for the core folks.

WojcikMike: Yeah I don't like it either, I asked number of people how to obsolete gracefully class that was changed to internal and the options were:
* rename the leftover class to something else, and keep the original type with obsolete
* move the leftover class to other namespace, and keep the original type with obsolete

Both of those solutions have pros and cons. What is more i don't have any good name for the namespace. So I am happy to adjust this change to something else that will allow us to achieve the goal

src/Tests/API/APIApprovals.ApproveAzureStorageQueueTransport.approved.txt
WojcikMike: Done.

src/Transport/AzureMessageQueueCreator.cs
SeanFeldman: Discussed with @WojcikMike and we decided to apply the following:

* Keep obsoleted stuff in the original namespace.
* Anything that is public and needs to be easily discovered by users in `NServiceBus` namespace.
* The rest in Project.Namespace (fe: `NServiceBus.AzureStorageQueues`)

WojcikMike:
After discussing with @SeanFeldman yesterday I introduced all outlined things.

I also addressed things that came out during conversation with @SimonCropp, @Scooletz and @andreasohlund 

Please review and merge

src/Transport/Dispatcher.cs
```diff
@@ -15,7 +16,7 @@
 
     class Dispatcher : IDispatchMessages
     {
-        public Dispatcher(CreateQueueClients createQueueClients, MessageWrapperSerializer messageSerializer, QueueAddressGenerator addressGenerator, AzureStorageAddressingSettings addressing)
+        public Dispatcher(AzureStorageQueues.CreateQueueClients createQueueClients, MessageWrapperSerializer messageSerializer, QueueAddressGenerator addressGenerator, AzureStorageAddressingSettings addressing)
```
Scooletz: Why `Dispatcher` requires a full name of `CreateQueueClients`?

Scooletz:
I'm :+1: with all the changes above. @SeanFeldman could you review it and merge if everything is ok? You had much more involvement in this, hence imo you'd be the one accepting it.

PR #91, opened 2016-04-19
Refactor Dispatcher to send messages concurrently

timbussmann:
changing Dispatcher to execute sends concurrently instead of dispatching them sequentially. This results in batched sends being faster due to better usage of parallel http requests.

Tests showed that the `WhenAll` is can be up to 3x faster than the sequential code. Note that this heavily depends on the `ServicePointManager.DefaultConnectionLimit`. The sequential approach has the side-effect of easier request reuse compared to concurrent requests which have to open new connections when the connection limit is too high.

Also switched to using `ExistsAsync`. Async all the way.

I think there are valid reasons for both the previous approach (foreach-await) and the suggested code (WhenAll). The later should outperform the first when sending multiple messages from a handler. It's hard to name a threshold when a concurrent dispatch (to azure) is faster than the sequential approach since it also heavily depends whether it has to open a new connection or reuse an existing one. Another good point came from @Scooletz: When hitting request limitations the sequential approach has to handle less 50X reponses and retries than when sending messages concurrently. Similar is for Queue existence checks: when multiple messages go to the same non-existent queue, callers will receive an `AggregateException` containing exceptions for each message in the batch compared to a single one.

@Particular/azure-storage-queues-maintainers please review.

### DIFF
--- a/src/Transport/Dispatcher.cs
+++ b/src/Transport/Dispatcher.cs
@@ -30,10 +30,13 @@
                 throw new Exception("The Azure Storage Queue transport only supports unicast transport operations.");
             }
 
+            var sends = new List<Task>(outgoingMessages.UnicastTransportOperations.Count());
             foreach (var unicastTransportOperation in outgoingMessages.UnicastTransportOperations)
             {
-                await Send(unicastTransportOperation).ConfigureAwait(false);
+                sends.Add(Send(unicastTransportOperation));
             }
+
+            await Task.WhenAll(sends).ConfigureAwait(false);
         }
 
         async Task Send(UnicastTransportOperation operation)
@@ -48,7 +51,7 @@
             var q = addressGenerator.GetQueueName(queue.QueueName);
             var sendQueue = sendClient.GetQueueReference(q);
 
-            if (!Exists(sendQueue))
+            if (!await ExistsAsync(sendQueue).ConfigureAwait(false))
             {
                 throw new QueueNotFoundException
                 {
@@ -80,6 +83,7 @@
             }
 
             var rawMessage = SerializeMessage(operation, timeToBeReceived);
+
             try
             {
                 await sendQueue.AddMessageAsync(rawMessage, timeToBeReceived, null, null, null).ConfigureAwait(false);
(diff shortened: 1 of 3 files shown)


src/Transport/Dispatcher.cs
```diff
@@ -30,10 +30,13 @@ public Dispatcher(CreateQueueClients createQueueClients, MessageWrapperSerialize
                 throw new Exception("The Azure Storage Queue transport only supports unicast transport operations.");
             }
 
+            var sends = new List<Task>(outgoingMessages.UnicastTransportOperations.Count());
```
timbussmann: blergh, this makes me think `outgoingMessages.UnicastTransportOperations` should be an `IList`

SeanFeldman: @danielmarbach also mentioned that iterating over IList allocates memory anyways

timbussmann: not saying it has a direct performance impact, but since it's not a parameter, I don't see any downside providing an IList instead of an enumerable.

danielmarbach: @timbussmann @boblangley and I discussed this as well. We postponed it because we didn't want to break downstreams again

timbussmann: yep can totally see that argument. It was just a note. Something we may consider for the next major. :+1:

Scooletz: If you want no allocations beside the collection, create an array and use `for`.

src/Transport/Dispatcher.cs
```diff
@@ -93,10 +97,10 @@ public Dispatcher(CreateQueueClients createQueueClients, MessageWrapperSerialize
             }
         }
 
-        bool Exists(CloudQueue sendQueue)
+        Task<bool> ExistsAsync(CloudQueue sendQueue)
         {
             var key = sendQueue.Uri.ToString();
-            return rememberExistence.GetOrAdd(key, keyNotFound => sendQueue.Exists());
+            return rememberExistence.GetOrAdd(key, keyNotFound => sendQueue.ExistsAsync());
```
Scooletz: Nice.

src/Transport/Dispatcher.cs
timbussmann: not sure wether `ToArray()` would really be the better choice here? ToArray also seems to new up objects while Count() is able to ICollection.Count property. Seems to be microptimization, but I'm happy to apply some changes if somebody can point out the best solution.

Scooletz: I meant:
```
var sends = new Task[](outgoingMessages.UnicastTransportOperations.Count());
```

timbussmann: @Scooletz but then you still need to iterate over outgoingMessages.UnicastTransportOperations to dispatch the messages + maintaining an index for the array access?

PR #103, opened 2016-04-25
Configure JSON as default serializer

### LINKED ISSUE #102: Use `SetDefault` to override serialization back to JSON

Although setting serializer has been greatly simplified, we can override the `SerializerDefinition` using the settings method `.SetDefault`. This will remove the configuration burden from end users, simplifying the upgrade guide as well.

The `Azure Service Bus Transport` setting:
https://github.com/Particular/NServiceBus.AzureServiceBus/commit/b08be0deb259bde5626e2037e999d51189945932#diff-abc21d576eebe5908bcca158905c4878R19

This issue requires changes in the docs as well:
https://github.com/Particular/docs.particular.net/edit/master/nservicebus/upgrades/asq-6to7.md

timbussmann:
resolves #102 

* Configures JSON as the default serializer if no specific serializer has been configured.
* Uses the configured Serializer as the serializer and **deserializer** for the ASQ message wrapper. This means:
 * ASQ does not support multi-deserialization for message wrappers (because we can't determine the content-type and because the API is internal to the core). But it supports multi-deserialization for the message body itself
 * All ASQ endpoints need to have the same message wrapper serialization configured
   * If the wrapper message uses the same serializer as the message body, no additional config is required
    * If the wrapper uses a different serializer, it needs to configure it using the `SerializeMessageWrapperWith` API.

WIP - do not merge

todo:
* [x] refactor tests
* [x] check exception messages on incompatible serializer configs in multi-deserialization systems.
* [x] doco PR


SeanFeldman:
@timbussmann will this PR include the changes we've discussed with @Scooletz (option to set up MessageWrapper serializer explicitly)?

timbussmann:
> @timbussmann will this PR include the changes we've discussed with @Scooletz (option to set up MessageWrapper serializer explicitly)?

yes, the option is still the `SerializeMessageWrapperWith` API, but unless you're running a multi-serializer system, you won't need to use it since it will default to the default-serializer (which also is set to JSON)

src/Tests/API/APIApprovals.ApproveAzureStorageQueueTransport.approved.txt
```diff
@@ -84,6 +84,19 @@
     {
         public IsHostedIn() { }
     }
+    public class MessageWrapper : NServiceBus.IMessage
```
timbussmann: this is only required for the acceptance test. Not sure the test is worth it to make this public?

timbussmann:
@Particular/azure-storage-queues-maintainers please review

src/Tests/API/APIApprovals.ApproveAzureStorageQueueTransport.approved.txt
WojcikMike: We had a conversation with @Scooletz and @SimonCropp and Simon raised that some serializers require for Message to be public. So in the end we agreed to change this to public.

so :+1: for that change

src/Transport/EnvelopeDeserializationFailed.cs
```diff
@@ -7,7 +7,7 @@ namespace NServiceBus.AzureStorageQueues
     class EnvelopeDeserializationFailed : SerializationException
     {
         public EnvelopeDeserializationFailed(CloudQueueMessage message, Exception ex)
-            : base("Failed to deserialize message envelope", ex)
+            : base($"Failed to deserialize message envelope for message with id {message.Id}. Make sure the configured serializer is used across all endpoints or configure the message wrapper serializer for this endpoint using the `SerializeMessageWrapperWith` extension on the transport configuration. Please refer to the Azure Storage Queue Transport configuration documentation for more details.", ex)
```
SeanFeldman: :+1: too bad we can't have permalinks. It would make this exception message pure gold.

src/Transport/MessageEnvelopeUnwrapper.cs
```diff
@@ -18,19 +18,12 @@ public MessageWrapper Unwrap(CloudQueueMessage rawMessage)
...
-                {
-                    throw new SerializationException("Failed to deserialize message with id: " + rawMessage.Id);
-                }
+                m = messageWrapperSerializer.Deserialize(stream);
             }
 
             if (m == null)
             {
-                throw new SerializationException("Failed to deserialize message with id: " + rawMessage.Id);
+                throw new SerializationException("Message is null");
```
SeanFeldman: Would it be more descriptive to say: "Failed to deserialize a message. Reason: message is null." ?

timbussmann: this exception will be wrapped in the `EnvelopeDeserializationFailed` exception which starts with "Failed to deserialize message envelope for message with id {message.Id}", I think that's ok?

SeanFeldman: :+1: perfect. Ignore my... ignorance :)

timbussmann:
doco PR raised here: https://github.com/Particular/docs.particular.net/pull/1457

PR #106, opened 2016-04-26
use timespans on config API

SimonCropp:


SimonCropp:
@WojcikMike thoughts?

WojcikMike:
I like it, changes don't require any deprecation as this is brand new

PR #107, opened 2016-04-26
Change location of binaries folder

WojcikMike:
Connects to Particular/PlatformDevelopment#723

Moving binaries 1 level up in a hierachy
The reason for that is to allign with others repositories where binaries is on the same level as src folder

### DIFF
--- a/src/Transport/NServiceBus.AzureStorageQueues.csproj
+++ b/src/Transport/NServiceBus.AzureStorageQueues.csproj
@@ -24,11 +24,11 @@
     <DebugSymbols>true</DebugSymbols>
     <DebugType>full</DebugType>
     <Optimize>false</Optimize>
-    <OutputPath>..\binaries\</OutputPath>
+    <OutputPath>..\..\binaries\</OutputPath>
     <DefineConstants>DEBUG;TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
-    <DocumentationFile>..\binaries\NServiceBus.Azure.Transports.WindowsAzureStorageQueues.xml</DocumentationFile>
+    <DocumentationFile>..\..\binaries\NServiceBus.Azure.Transports.WindowsAzureStorageQueues.xml</DocumentationFile>
     <NoWarn>1591, 0618</NoWarn>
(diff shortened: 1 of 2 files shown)


WojcikMike:
@Particular/azure-storage-queues-maintainers please review and merge

melkio:
All "azure" related repos have the same convention. Do we have to update all of them?

WojcikMike:
@andreasohlund told me that it is probably better to update as keeping binaries in the src directory doesn't make sense. But as I am not maintainer I am not in the position to say, that is the only way to do it.

melkio:
My question was only to know if we have to create a specific issue for each azure related repos. That's all :)

src/Transport/NServiceBus.AzureStorageQueues.csproj
```diff
@@ -24,23 +24,23 @@
...
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <DebugType>pdbonly</DebugType>
     <Optimize>true</Optimize>
-    <OutputPath>..\binaries\</OutputPath>
+    <OutputPath>..\..\binaries\</OutputPath>
     <DefineConstants>TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
-    <DocumentationFile>..\binaries\NServiceBus.Azure.Transports.WindowsAzureStorageQueues.xml</DocumentationFile>
+    <DocumentationFile>..\..\binaries\NServiceBus.Azure.Transports.Win  dowsAzureStorageQueues.xml</DocumentationFile>
```
SimonCropp: why the extra spaces?

WojcikMike: Thanks for catching that @SimonCropp fixed

WojcikMike:
@melkio yes I think we should.

Scooletz:
@melkio My understanding is that @Particular/azure-storage-queues-maintainers will support this, but applying in other repos is their's maintainers responsibility.

WojcikMike:
@melkio I created Issues in:
* AzureStoragePersistence
* AzureServiceBus
* AzureBlobStorageDataBus
* AzureCloudService

Let me know if I missed something

SeanFeldman:
:+1:

PR #108, opened 2016-04-26
Changed circuitbreaker success and failure calls

timbussmann:
Connects to Particular/PlatformDevelopment#723

* moved success call right after successful receive, since there is no need to wait until the limiter has been passed and messages have been dispatched
* do no longer arm (.fail) the circuit breaker when message processing fails. This was basically useless since the it will always call success after polling the queue and therefore immediately disarm the CB again. If message handling fails this constantly logs "circuit breaker is now armed" but never actually triggering. According to @yvesgoeleven it's intended to not trigger on repeated message processing failures, so it seems like we can remove it there. The only reason the circuit breaker now triggers is in scenarios where ASQ can't be read over the configured time period.

@Particular/azure-storage-queues-maintainers please review

### DIFF
--- a/src/Transport/MessagePump.cs
+++ b/src/Transport/MessagePump.cs
@@ -118,7 +118,8 @@
                 try
                 {
                     var retrieved = await messageReceiver.Receive(cancellationTokenSource.Token).ConfigureAwait(false);
-
+                    circuitBreaker.Success();
+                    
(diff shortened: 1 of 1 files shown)


Scooletz:
I see the reasoning and the changes follow it, but would like to have it merged by @yvesgoeleven

PR #114, opened 2016-04-29
Logical mapping of the default account fixed.

Scooletz:
The default account mapping was failing if logical-connection_string mapping was enabled. This fixes it.

### DIFF
--- a/src/AcceptanceTests/ConfigureAzureStorageQueueTransport.cs
+++ b/src/AcceptanceTests/ConfigureAzureStorageQueueTransport.cs
@@ -27,7 +27,9 @@ public class ConfigureEndpointAzureStorageQueueTransport : IConfigureEndpointTes
         configuration.UseTransport<AzureStorageQueueTransport>()
             .ConnectionString(connectionString)
             .MessageInvisibleTime(TimeSpan.FromSeconds(5))
-            .SerializeMessageWrapperWith<JsonSerializer>();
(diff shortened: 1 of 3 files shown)


Scooletz:
Ping @Particular/azure-storage-queues-maintainers @WojcikMike

timbussmann:
from the conversation I had today with @melkio , wouldn't this break backward compatibility with v6 endpoints?

timbussmann:
here some scenarios as I understood them from the call today:

* v7 endpoint receives v6 message which contains a connection string
  * after the transport, the connection string shouldn't be visible on the headers, regardless of the `UseAccountNamesInsteadOfConnectionStrings()` option.
     * since v7 doesn't support multi-account yet, the incoming reply to address should be the same connection string anyway. If not, should we just throw?
* v7 endpoint running without `UseAccountNamesInsteadOfConnectionStrings()`: outgoing sends should contain the connection string for backward compatibility -> I don't see a way to test this in a ATT?
* v7 endpoint running with `UseAccountNamesInsteadOfConnectionStrings()`: outgoing sends should not contain the connection string. no account name / alias is required since we don't support multi-accounts as of now. -> also not testable in ATT?

@melkio please verify :point_up:

Scooletz:
@timbussmann 
The fix ensures that when mapping is enabled queues in the default account can be used without prefixes and are mapped correctly. I see no way of breaking the backward compatibility. Could you provide a scenario for this?

PR #115, opened 2016-04-29
Account name mapping tests

### LINKED ISSUE #63: Review configuration API
Configuration API should be more similar to ASB config API, to improve users experience across Azure transports.
Namespaces have to be reviewed to make config API easily discoverable

### LINKED ISSUE #95: purpose of IAzureStoragePartitioningSettings?
it only has one implementation. and i see no way passing an instance into the api. and the only time it is exposed an instance of AzureStorageAddressingSettings is returned

Scooletz:

Following tests should be provided:
- send in the same account without connection string mapping
- send in the same account with connection string mapping
- ~~send across accounts without connection string mapping~~
- send across accounts with connection string mapping

Additionally:
- closes #95 
- closes #63


src/AcceptanceTests/Configuration/When_using_mapping_to_account_names.cs
```diff
@@ -0,0 +1,97 @@
...
+            public Receiver()
+            {
+                EndpointSetup<DefaultPublisher>();
+                CustomEndpointName(ReceiverName);
+                AuditTo(AuditName);
+            }
+
+            public class Handler : IHandleMessages<MyCommand>
+            {
+                public Context Context { get; set; }
```
andreasohlund: Use ctor injection so that we're prepared for the day when property injection won't be supported ootb

timbussmann:
removed cross-account sends from the list since currently not supported in v7

timbussmann:
here some scenarios as I understood them from the call today:

* v7 endpoint receives v6 message which contains a connection string
  * after the transport, the connection string shouldn't be visible on the headers, regardless of the `UseAccountNamesInsteadOfConnectionStrings()` option.
     * since v7 doesn't support multi-account yet, the incoming reply to address should be the same connection string anyway. If not, should we just throw?
* v7 endpoint running without `UseAccountNamesInsteadOfConnectionStrings()`: outgoing sends should contain the connection string for backward compatibility -> I don't see a way to test this in a ATT?
* v7 endpoint running with `UseAccountNamesInsteadOfConnectionStrings()`: outgoing sends should not contain the connection string. no account name / alias is required since we don't support multi-accounts as of now. -> also not testable in ATT?

@melkio please verify :point_up:

Scooletz:
A remark:
The cross-account sends are being discussed right now. This PR will be updated accordingly.

Scooletz:
I covered cross account sends with tests. Now one can work with them on multi-account support. This does not tell if we ship with multi-account in beta stage or not.

Scooletz:
/cc @WojcikMike @timbussmann

WojcikMike:
Looks good to me

src/Tests/API/APIApprovals.ApproveAzureStorageQueueTransport.approved.txt
```diff
@@ -156,8 +151,7 @@ namespace NServiceBus
     }
     public class static AzureStorageTransportAddressingExtensions
     {
-        public static NServiceBus.AzureStorageAddressingSettings Addressing(this NServiceBus.TransportExtensions<NServiceBus.AzureStorageQueueTransport> config) { }
-        public static NServiceBus.IAzureStoragePartitioningSettings Partitioning(this NServiceBus.AzureStorageAddressingSettings addressingSettings) { }
+        public static NServiceBus.TransportExtensions<NServiceBus.AzureStorageQueueTransport> UseAccountNamesInsteadOfConnectionStrings(this NServiceBus.TransportExtensions<NServiceBus.AzureStorageQueueTransport> config, string defaultConnectionStringName, System.Collections.Generic.Dictionary<string, string> name2connectionString = null) { }
```
melkio: Using a dictionary could be not so meaningful. Is it possible to use a lambda to create internally the dictionary and to express better what the user has to do?

Scooletz: I need some code proposal. Are you referring to using some kind of previously existing `.AddAccount` but in a lambda?

Scooletz: We agreed that a small builder will be a better approach:
```
public static NServiceBus.TransportExtensions<NServiceBus.AzureStorageQueueTransport> UseAccountNamesInsteadOfConnectionStrings(this NServiceBus.TransportExtensions<NServiceBus.AzureStorageQueueTransport> config, string defaultConnectionStringName, Action<AccountMapping> mapAdditionalAccounts) 
{
//... 
}
```

src/Tests/API/APIApprovals.ApproveAzureStorageQueueTransport.approved.txt
```diff
@@ -1,4 +1,39 @@
-﻿namespace NServiceBus.Azure.Transports.WindowsAzureStorageQueues
+﻿namespace NServiceBus
+{
+    
+    public sealed class AccountMapping
```
melkio: Why `public`?

src/Tests/API/APIApprovals.ApproveAzureStorageQueueTransport.approved.txt
```diff
@@ -1,4 +1,39 @@
...
+        protected override NServiceBus.Transports.TransportInfrastructure Initialize(NServiceBus.Settings.SettingsHolder settings, string connectionString) { }
+    }
+    public class static AzureStorageTransportAddressingExtensions
+    {
+        public static NServiceBus.TransportExtensions<NServiceBus.AzureStorageQueueTransport> UseAccountNamesInsteadOfConnectionStrings(this NServiceBus.TransportExtensions<NServiceBus.AzureStorageQueueTransport> config, string defaultConnectionStringName, System.Action<NServiceBus.AccountMapping> map) { }
+    }
+    public class static AzureStorageTransportExtensions
+    {
+        public static NServiceBus.TransportExtensions<NServiceBus.AzureStorageQueueTransport> BatchSize(this NServiceBus.TransportExtensions<NServiceBus.AzureStorageQueueTransport> config, int value) { }
+        public static NServiceBus.TransportExtensions<NServiceBus.AzureStorageQueueTransport> ConnectionString(this NServiceBus.TransportExtensions<NServiceBus.AzureStorageQueueTransport> config, string value) { }
```
melkio: Why does transport expose this configuration API? It's already provided by `TransportExtension<T>`

src/Transport/Config/Extensions/AzureStorageTransportAddressingExtensions.cs
```diff
@@ -1,25 +1,39 @@
...
 {
+    using System;
+    using System.Collections.Generic;
     using Configuration.AdvanceExtensibility;
 
     public static class AzureStorageTransportAddressingExtensions
     {
-        public static AzureStorageAddressingSettings Addressing(this TransportExtensions<AzureStorageQueueTransport> config)
+        public static TransportExtensions<AzureStorageQueueTransport> UseAccountNamesInsteadOfConnectionStrings(this TransportExtensions<AzureStorageQueueTransport> config,
+            string defaultConnectionStringName,
```
melkio: I'm thinking about the first parameter: why doesn't transport give only the lambda to configure all possible accounts? With two different methods if needed (one for default, one the the others accounts).

Except for this point, this new implementation is fine by me.

src/Tests/API/APIApprovals.ApproveAzureStorageQueueTransport.approved.txt
Scooletz: It's used in the `Action<AccountMapping>` for the `UseNames...` method.

src/Tests/API/APIApprovals.ApproveAzureStorageQueueTransport.approved.txt
Scooletz: It's highly unlikely that my change caused it. This part has been moved in the API file when I was doing the compare. Let me check it. I assume you meant 'ConnectionString` method?

src/Transport/Config/Extensions/AzureStorageTransportAddressingExtensions.cs
Scooletz: First, you probably don't want to configure the connection string first on the general level and later on pass it again to map the default. Second, keeping the builder part minimal (one method only) and separated from setting just a name for the default works better imo. Users who want to map additional accounts, will do it. Users who don't, will just use one obligatory parameter.
On the other hand, these who don't use cross-account sends, don't need the name, do they?

Could you sketch the API you described? A piece of code would be great.

src/Tests/API/APIApprovals.ApproveAzureStorageQueueTransport.approved.txt
Scooletz: It looks like this `ConnectionString` can be removed because it's provided one the core level right now. I won't spend time on digging it up. I'm removing it as it uses the exactly same core.

src/AcceptanceTests/Sending/When_dispatching_to_other_account.cs
```diff
@@ -62,9 +62,7 @@ public Endpoint()
                 EndpointSetup<DefaultServer>(configuration =>
                 {
                     configuration.UseTransport<AzureStorageQueueTransport>()
-                        .Addressing()
-                        .Partitioning()
-                        .AddStorageAccount(AnotherAccountName, Transports.Default.Settings.Get<string>("Transport.ConnectionString"));
+                        .UseAccountNamesInsteadOfConnectionStrings("default", _ => _.MapAccount(AnotherAccountName, Transports.Default.Settings.Get<string>("Transport.ConnectionString")));
```
dvdstelt: Isn't the idea of _ that it's a parameter you're never going to use? But here I see it is in fact used. Why is it being called _?

Scooletz: That's the first time I'm reading about `_` being used for an unused parameter. I've googled it and it looks like it's a common way in C# to mark a param as unused. I wasn't aware of this notation :) I've pushed a renaming commit.
I think I've seen somewhere a proposal for Roslyn to use `_` as a default lambda param, but I can't find it right now.

src/Transport/Config/Extensions/AzureStorageTransportAddressingExtensions.cs
melkio: Something like:

```
   extensions.UseAccountNamesInsteadOfConnectionStrings(c => 
   {
      c.MapLocalAccount("defaultName);
      c.MapAccount("alias1", "connectionString1");
      c.MapAccount("alias2", "connectionString2");
   });
```

PR #118, opened 2016-05-11
An empty lambda is allowed for secure sends in one account

Scooletz:
After changing the API, passing the name for the default account is no longer required from the API point of view. Unfortunately it breaks UX when using one account as it requires to call the mapping method for no reason. This PR fixes this behavior allowing to skip the mapping when using only one account.
Tests have been provided.

### DIFF
--- a/src/AcceptanceTests/Configuration/Should_accept_no_mappings_at_all.cs
+++ b/src/AcceptanceTests/Configuration/Should_accept_no_mappings_at_all.cs
@@ -1,0 +1,113 @@
+﻿namespace NServiceBus.AcceptanceTests.WindowsAzureStorageQueues.Configuration
+{
+    using System;
+    using System.Linq;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using AcceptanceTesting.Support;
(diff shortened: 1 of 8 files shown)


dvdstelt:
What I understand is that with no mapping, we basically do this:

.UseAccountNamesInsteadOfConnectionStrings(_ => { })

Why not create an overload without a mapping?

```
public static TransportExtensions<AzureStorageQueueTransport> UseAccountNamesInsteadOfConnectionStrings(this TransportExtensions<AzureStorageQueueTransport> config)
{
    return UseAccountNamesInsteadOfConnectionStrings(config, _ => { });
}
```

dvdstelt:
Also, there's no documentation at all on AccountMapping, although there is documentation on `UseAccountNamesInsteadOfConnectionStrings`
http://docs.particular.net/nservicebus/azure-storage-queues/configuration#connection-strings-securing-connection-strings-to-storage-accounts

Should we not explain how it's used?

Scooletz:
I'll create the overload. As soon as this is merged, I'll update this section. I need the fixes published first.

PR #130, opened 2016-06-30
QueueReceiver only allocates once for empty receives and returns a list

danielmarbach:
instead of an enumerable to save allocations

### DIFF
--- a/src/Transport/AzureMessageQueueReceiver.cs
+++ b/src/Transport/AzureMessageQueueReceiver.cs
@@ -54,15 +54,20 @@ namespace NServiceBus.AzureStorageQueues
             }
         }
 
-        internal async Task<IEnumerable<MessageRetrieved>> Receive(CancellationToken token)
+        internal async Task<List<MessageRetrieved>> Receive(CancellationToken token)
         {
             var rawMessages = await azureQueue.GetMessagesAsync(BatchSize, MessageInvisibleTime, null, null, token).ConfigureAwait(false);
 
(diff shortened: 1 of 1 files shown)


src/Transport/AzureMessageQueueReceiver.cs
```diff
@@ -54,15 +54,20 @@ public AzureMessageQueueReceiver(MessageEnvelopeUnwrapper unwrapper, CloudQueueC
...
         }
 
-        internal async Task<IEnumerable<MessageRetrieved>> Receive(CancellationToken token)
+        internal async Task<List<MessageRetrieved>> Receive(CancellationToken token)
         {
             var rawMessages = await azureQueue.GetMessagesAsync(BatchSize, MessageInvisibleTime, null, null, token).ConfigureAwait(false);
 
             var messageFound = false;
-            var messages = new List<MessageRetrieved>();
+            var messages = noMessagesFound;
```
Scooletz: What if we assinged `null` in here:
```
List<MessageRetrieved> messages = null;
```
and return
``` 
returns messages ?? noMessagesFound;
```
IMO this is better as you won't accidentally introduce a change that changes `noMessagesFound`

danielmarbach: Good for me

Scooletz: Will you issue an update?

danielmarbach:
@Scooletz pushed

PR #132, opened 2016-06-30
Replaced EnvelopeDeserializationFailed exception with SerializationException

danielmarbach:
A custom private exception type `EnvelopeDeserializationFailed` which is not interecepted does not make sense. It can be replaced with a general purpose `Exception`.


dvdstelt:
What is the reason for this PR?

SeanFeldman:
Unneeded custom exception @dvdstelt

Issue #119, opened 2016-05-12
Secure connections strings configuration API is not clear


SeanFeldman:
Connects to https://github.com/Particular/PlatformDevelopment/issues/871

Going through configuration API and reading ["Securing connections strings to storage accounts"](http://docs.particular.net/nservicebus/azure-storage-queues/configuration#connection-strings-securing-connection-strings-to-storage-accounts) I've realized that I don't understand how the API should be used. The 2 almost identical code snippets are very confusing:

```c#
var transport = endpointConfiguration.UseTransport<AzureStorageQueueTransport>();
transport
    .ConnectionString("default_connection_string")
    .UseAccountNamesInsteadOfConnectionStrings(mapping =>
    {
        mapping.MapLocalAccount("default");
        mapping.MapAccount("another","another_connection_string");
    });
```
and
```c#
var transport = endpointConfiguration.UseTransport<AzureStorageQueueTransport>();
transport
    .ConnectionString("another_connection_string")
    .UseAccountNamesInsteadOfConnectionStrings(mapping =>
    {
        mapping.MapLocalAccount("another");
        mapping.MapAccount("default", "default_connection_string");
    });
```
The questions I had were:

1. Why the two identical code snippets?
1. What's the affect on configuration if `.ConnectionString()` is used? Is that a `default` connections string?
1. What is `MapLocalAccount()`?
1. How `.ConnectionString()` and `.UseAccountNamesInsteadOfConnectionStrings()` with mapping affect default connection string?
1. How `mapping.MapAccount("default", "default_connection_string");` relates to `ConnectionString()`?

Scooletz:
Let me try to answer your questions. But first, let me remark that maybe this documentation piece should be moved to http://docs.particular.net/nservicebus/azure-storage-queues/multi-storageaccount-support
 
> 1. Why the two identical code snippets?

Every snippet represents a separate endpoint using a different account for its input queue.

> 2. What's the affect on configuration if .ConnectionString() is used? Is that a default connections string?

Yes, that's the default connection string.

> 3. What is MapLocalAccount(name)

This maps the default, local connection string with the name, so when sending to another account, the name will be used for the `ReplyTo` header.

> 4. How .ConnectionString() and .UseAccountNamesInsteadOfConnectionStrings() with mapping affect default connection string?

Mapping does not change the connection string value. If you set the connection string with `.ConnectionString()` it is set. Mapping is only for cross account sends.

[...]

PR #134, opened 2016-07-05
Align connection string mapping to ASB API

Scooletz:
Connects to https://github.com/Particular/PlatformDevelopment/issues/871

On the basis of https://github.com/Particular/NServiceBus.AzureStorageQueues/issues/119

##### API after the change

```
cfg.UseTransport<AzureStorageQueueTransport>()		
   .UseAccountNamesInsteadOfConnectionStrings()
   .DefaultAccountName(DefaultConnectionStringName)
   .AccountRouting()
   .AddAccount(AnotherConnectionStringName, anotherConnectionString);
```

##### Validation change

The PR changes the place where an exception can be thrown when using `UseAccountNamesInsteadOfConnectionStrings`. Currently the method only marks up the configuration as one using account names. The application is done during call to `AzureStorageQueueInfrastructure.ConfigureReceiveInfrastructure`. This changes the place when the validation of providing the default account name occurs. That's the only way I see it can be done if we want to provide a truly fluent interface with `UseAccountNamesInsteadOfConnectionStrings` possible to be called any time during configuration.

### DIFF
--- a/src/AcceptanceTests/Configuration/Should_accept_no_mappings_at_all.cs
+++ b/src/AcceptanceTests/Configuration/Should_accept_no_mappings_at_all.cs
@@ -7,7 +7,9 @@
     using AcceptanceTesting.Support;
     using Azure.Transports.WindowsAzureStorageQueues.AcceptanceTests;
     using EndpointTemplates;
+    using Features;
     using NUnit.Framework;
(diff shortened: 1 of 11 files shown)


Scooletz:
Maintainer involved in the ASB, please review the alignment of the API (the implementation is slightly different).

PR #127, opened 2016-06-10
Transport error handling

Scooletz:
Connects to https://github.com/Particular/PlatformDevelopment/issues/832

This PR beside the alignment to the new recoverability extracts two implementations of `ReceiveStrategy`: `AtLeastOnce` and `AtMostOnce`. This clarifies a lot a different set of responsibilities which they have, when dispatching a message.

Notes:
- currently ASQ Transport uses only the `main serializer` to serialize/deserialize `MessageWrapper` and this cannot be changed. Because of this, one of the serializer tests is `Ignored` for now.
- an internal `SettingsHolder.Merge(...)` method is called via reflection to enable combining serializer settings with default settings (a'la `SerializationFeature`).

PR #137, opened 2016-07-22
[WIP] Use byte array instead of stream

danielmarbach:
See https://github.com/Particular/NServiceBus/pull/3956

Does not address all other release branch changes from core


Scooletz:
This is handled by #127 where the new seam is introduced.

PR #138, opened 2016-07-27
Pop receipt coherent handling

### LINKED ISSUE #126: Message deletion behaviour is not consistent
## Who's affected

- Any ASQ users that run with a single consumer (with concurrency of 1) and later either scaling out or increasing concurrency level.

## Symptoms

- With a single consumer, message can be deleted after `VisibilityTimeout` elapses. With multiple consumers, this is not going to be possible.

## Analysis

This is an issue with the MSFT ASQ library design. It provided a pop receipt for a message w/o connecting it to visibility timeout. I have [reported the issue](https://github.com/Azure/azure-storage-net/issues/285) and it's now a feature request. 

A possible workaround is to correlate between time when message was received and attempted for deletion [suggested by the ASQ team](https://github.com/Azure/azure-storage-net/issues/285#issuecomment-223887791).

Scooletz:
Fixes #126

### DIFF
--- a/src/Transport/MessageRetrieved.cs
+++ b/src/Transport/MessageRetrieved.cs
@@ -48,9 +48,11 @@
         {
             if (rawMessage.NextVisibleTime != null)
             {
-                if (rawMessage.NextVisibleTime.Value < DateTimeOffset.Now)
+                var visibleIn = rawMessage.NextVisibleTime.Value - DateTimeOffset.Now;
+                if (visibleIn < TimeSpan.Zero)
                 {
-                    throw new LeaseTimeoutException();
+                    var visibilityTimeoutExceededBy = -visibleIn;
+                    throw new LeaseTimeoutException(rawMessage, visibilityTimeoutExceededBy);
                 }
             }
         }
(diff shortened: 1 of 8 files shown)


Scooletz:
@SeanFeldman could you review it as you're the author of #126

src/Transport/MessageRetrieved.cs
```diff
@@ -67,4 +82,11 @@ public MessageWrapper Unwrap()
         CloudQueue azureQueue;
         CloudQueueMessage rawMessage;
     }
+
+    class LeaseTimeoutException : Exception
+    {
+        public LeaseTimeoutException() : base("Pop receipt is invalid as it exceeded the next visibile time.")
```
SeanFeldman: Should the exception capture the original Visibility time which caused the exception to be raised? Or the message ID that caused this exception to be raised? Thinking as a customer: reading the log that says 

> WARN: Dispatching the message took longer than a visibility timeout. The message will reappear in the queue and will be obtained again. Exception: Pop receipt is invalid as it exceeded the next visibile time.

could be hard to troubleshoot

SeanFeldman: "visibile" -> visible

SeanFeldman:
Looks good @Scooletz 
2 minor items and done.

src/Transport/MessageRetrieved.cs
Scooletz: Cannot agree more @SeanFeldman I'll add the estimated visibility time and the message identifier as well.

Scooletz:
Ping @SeanFeldman

SeanFeldman:
:+1: merged

PR #143, opened 2016-08-18
Poison messages handling

Scooletz:
Addresses #142

Currently, when a malformed message is retrieved and cannot be deserialized properly ASQ handles it poorly. This PR addresses this need by explicitly catching any serialization related problems and putting a problematic message directly to the error queue.

PR #147, opened 2016-08-29
Unwrapping raw messages is done concurrently

danielmarbach:
I was reviewing https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/143 and saw the following behavior change:

## Before 143

Unwrapping happens on the worker thread pool concurrently (deserialization is currently a CPU bound operation)

## After 143

Unwrapping happens sequentially on the receive thread per message

When I moved the unwrapping into the receive I found out that it has significant perf boost when we can free up the receive thread. See https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/79 

This PR goes back to concurrent unwrapping and aligns cloud queue operations under the umbrella of the `MessageRetrieved ` class.

### DIFF
--- a/src/Transport/MessageRetrieved.cs
+++ b/src/Transport/MessageRetrieved.cs
@@ -1,6 +1,7 @@
 ﻿namespace NServiceBus.AzureStorageQueues
 {
     using System;
+    using System.Runtime.Serialization;
     using System.Threading.Tasks;
     using Azure.Transports.WindowsAzureStorageQueues;
     using Microsoft.WindowsAzure.Storage;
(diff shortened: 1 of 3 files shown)


src/Transport/MessageRetrieved.cs
```diff
@@ -1,30 +1,52 @@
...
         /// <summary>
+        /// Unwraps the raw message body.
+        /// </summary>
+        /// <exception cref="SerializationException">Thrown when the raw message could not be unwrapped. The raw message is automatically moved to the error queue before this exception is thrown.</exception>
+        /// <returns>The actual message wrapper.</returns>
+        public async Task<MessageWrapper> Unwrap()
+        {
+            try
+            {
+                return unwrapper.Unwrap(rawMessage);
```
Scooletz: I was considering providing a caching fields for the unwrapped message. A message is unwrapped in one place only, so this wouldn't be optimization at all. YAGNI.

PR #149, opened 2016-08-31
[WIP] Changed the ToTransportAddress behavior to only used relevant fields

SzymonPobiega:
The previous version with ToString would include all instance properties (if provided) which is not something we want in the address.

### DIFF
--- a/src/Transport/AzureStorageQueueInfrastructure.cs
+++ b/src/Transport/AzureStorageQueueInfrastructure.cs
@@ -2,6 +2,7 @@
 {
     using System;
     using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


Scooletz:
@SzymonPobiega I'm about to issue PRs for bumping up current release to RC. When your PR is supposed to be merged? Is this `v6` RC or later?

SzymonPobiega:
The core PR is going to be postponed to next major so closing this one.

SzymonPobiega:
Here's the equivalent based on the current core: https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/151

PR #150, opened 2016-09-01
Update Core Rc1

Scooletz:
Connects to Particular/V6Launch#3

PR #151, opened 2016-09-01
Made transport address translation explicit

SzymonPobiega:
The `ToString` was meant to be a debugging helper. It incudes instance properties which you don't want to have in the default form in the transport address.


SzymonPobiega:
@Particular/azure-storage-queues-maintainers please review.

Scooletz:
@SzymonPobiega The build has failed. Let me go through this as this happened for another PR as well.

Scooletz:
This requires rebuild after #150 is merged.

PR #152, opened 2016-09-01
Update to GitVersion 3.6.3

Scooletz:
Connects to Particular/V6Launch#3

### DIFF
--- a/GitVersion.yml
+++ b/GitVersion.yml
@@ -1,0 +1,7 @@
+assembly-versioning-scheme: Major
+next-version: 7.0
+branches:
+  develop:
+    tag: alpha
+  releases?[/-]:
+    tag: rc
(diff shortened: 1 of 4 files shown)


GitVersion.yml
```diff
@@ -0,0 +1,7 @@
+assembly-versioning-scheme: Major
+next-version: 7.0
+branches:
+  develop:
+    tag: alpha
+  releases?[/-]:
```
SeanFeldman: Is that needed?

Scooletz: As far as I understand, the file should be named in this way and `GitVersion` no longer uses branches for selecting a version number.

Scooletz:
This requires rebuild after #150 is merged.

PR #155, opened 2016-09-28
[WIP] Native delayed delivery

Scooletz:
Related to https://github.com/Particular/NServiceBus/issues/4085
Solves the following natively in ASQ: https://github.com/Particular/NServiceBus.Persistence.AzureStorage/issues/6
Elects single timeout poller when scaled out: https://github.com/Particular/NServiceBus/issues/2602

This PR provides a native support for `delayed delivery` in ASQ.

Summary:
- when a timeout delay is lower than `MaxVisibilityDelay`, a native ASQ mechanism `visibilityTimeout` is used
- when a timeout delay is greater than `MaxVisibilityDelay`, a `TimeoutEntity` is written to a dedicated table in a format that enables to recreate a message to be sent
-  `MaxVisibilityDelay = CloudQueueMessage.MaxTimeToLive - TimeSpan.FromDays(1)` - just to not expire timeouts just before dispatching
- a custom `TimeoutsPoller` is provided to query the store and dispatch messages to the target queues
- if a target queue does not exists (for any reason, like an endpoint disposal), the poller will place the message in the endpoints` queue
- a blob lease lock manager is implemented to run a single `TimeoutsPoller` for every endpoint, across its all instances ([original ASP issue](https://github.com/Particular/NServiceBus.Persistence.AzureStorage/issues/6))

Abandoned ideas:
- the initial idea of using `AppendOnlyBlob` was abandoned as it enables only 50,000 blocks. Creation of new blobs for a specific time period etc would be much more harder than managing a table + a blob based lock.

TODO:
- [x] handle properly exceptions in the poller, without stopping it
- [x] what about queues that are no longer there when timeout occurs? Should this timeout be deleted?
- [x] `AtLeastOnce` `AtMostOnce` should they be applied to the timeouts' dispatching as well?
- [x] do we need more tests? 
 - Lease manager
- [x] address `MessagePump` running for the main queue as well as satellites. It should run only for the main.


Scooletz:
Ping @andreasohlund 

What do you think?

andreasohlund:
I like the general idea

>a customer feature + a behavior are used to do not make the pump clumsy. It's much better to operate in the pipeline itself, where the dispatcher and the next() are available

This sounds off, lets get on a call and see what's missing since I believe that having to use a behavior is a sign that the transport seam might need adjustment

We talked about storing the > 7 days messages in table storage. Do you still believe that would work? (less rescheduling ops)

src/AcceptanceTests/Sending/When_delaying_messages_natively.cs
```diff
@@ -0,0 +1,157 @@
...
+                .Run();
+            Assert.True(context.WasCalled, "The message handler should be called");
+        }
+
+        [Test]
+        public async Task Should_resend_message_if_delay_did_not_pass()
+        {
+            var delay = TimeSpan.FromMinutes(1);
+
+            // remove the visibilitytimeout up to 5 times, which will make the message to be requeued again and again
```
SeanFeldman: What's the scenario you're testing here? Why would a single header be stripped?

src/Transport/AzureStorageQueueInfrastructure.cs
```diff
@@ -18,13 +20,22 @@ internal AzureStorageQueueInfrastructure(ReadOnlySettings settings, string conne
...
+            useNativeTimeouts = settings.GetOrDefault<bool>(WellKnownConfigurationKeys.NativeTimeouts);
+
+            var contraints = new List<Type>
+            {
+                typeof(DiscardIfNotReceivedBefore),
+                typeof(NonDurableDelivery)
+            };
+            if (useNativeTimeouts)
+            {
+                contraints.Add(typeof(DelayedDeliveryConstraint));
```
SeanFeldman: This would mean Saga timeouts will go through native timeouts as well causing anything more than 7 days to be reset to 7, wouldn't it?

src/AcceptanceTests/Sending/When_delaying_messages_natively.cs
Scooletz: I'm testing a reappearing message after some time. The easiest way to me, was just remove the header a few times making the message reappear again and again, being filtered out by the behavior till the time comes.
Maybe I need to rethink this approach or at least, rethink assertions.

src/Transport/AzureStorageQueueInfrastructure.cs
Scooletz: Yes, saga timeouts will go through this. No, they won't be reset. The Dispatcher adds a special header writing the time when a message should be dispatched. When the span is greater than 7 days, a message will reappear after 7 days and will be requeued for more.

Scooletz:
@andreasohlund 
>This sounds off, lets get on a call and see what's missing since I believe that having to use a behavior is a sign that the transport seam might need adjustment

Cool. Let's discuss it.

> We talked about storing the > 7 days messages in table storage. Do you still believe that would work? (less rescheduling ops)

Yes. I wanted to first finish the initial sketch. We can discuss it in relation to the first point you mentioned.

src/Transport/AzureStorageQueueInfrastructure.cs
SeanFeldman: Ah, hence the custom behaviour to hide the message if time has not elapsed yet.

src/Transport/NativeDelayDeliveryFeature.cs
```diff
@@ -0,0 +1,131 @@
...
+                    // if next in the past, dispatch now
+                    if (diff < TimeSpan.Zero)
+                    {
+                        return next();
+                    }
+
+                    // TODO: what with other constraints like sending a timeout with TimeSpan.FromMinutes(5) which should be expired after TimeSpan.FromMinutes(6) when not delivered. This should be copied to the new message
+
+                    var msg = context.Message;
+                    var outgoingMessage = new OutgoingMessage(msg.MessageId, msg.Headers, msg.Body);
```
SeanFeldman: What about the original message `DequeueCount`? It won't be preserved when a new messages is created to keep it hidden. Though `DequeueCount` is needed for native retries. Am I missing something? Thanks

src/Transport/AzureStorageQueueInfrastructure.cs
Scooletz: Yes. Using a behavior seemed to me a bit cleaner, but after @andreasohlund https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/155#issuecomment-250257302 I'm not that sure any longer :) I can move that to pump easily, but this is so behaviorish that I started with this approach.

src/Transport/NativeDelayDeliveryFeature.cs
Scooletz: You're right @SeanFeldman ! This looks like ASB mixed approach for storing this, doesn't it? Provide a custom header with the counter... On the other hand, once the message is ready to be dispatched it won't be sent with a non-null `visibilitytimeout`, so I don't have to preserve this counter, do I?

Scooletz:
One more thing that came to mind after writing all the answers. If we use tables to store timeouts we'll need to ensure that only one poller is retrieving messages from tables as there is no `lease` operation on a table row (there's a pop lease for messages). This could be fixed by using blobs and leases to ensure that only one timeout dispatcher is run, which goes back to https://github.com/Particular/NServiceBus/issues/2602 . I'm not saying it's wrong, just noticing that timeouts would require all the three storage services:
- queues :)
- tables - to store timeouts data
- blobs - to lease and ensure only one poller is run at the same time

src/Transport/NativeDelayDeliveryFeature.cs
SeanFeldman: Right. Since timeout to be delivered is no longer going to be hidden, we're good. Scratch my question :)

SeanFeldman:
Yeah, having it all on a message is probably a cleaner solution.
@andreasohlund would there be an alternative for the behaviour part?

Scooletz:
Let me ping this PR tomorrow. I'll have something new to offer I presume, using `AppendBlob`.

Scooletz:
@SeanFeldman @andreasohlund I've updated the description of this PR and provided an implementation based on Tables + blob lease lock.

andreasohlund:
Talked this through with @Scooletz => ❤️

Scooletz:
@yvesgoeleven thoughts?

SeanFeldman:
> Talked this through with @Scooletz => ❤️

@andreasohlund and you've recorded/summed it up somewhere?...

andreasohlund:
Not really: ❤️ is the best I can give at this stage :)

SeanFeldman:
@Scooletz the message based version is gone completely?

Scooletz:
No it isn't @SeanFeldman I've updated the PR description. There wasn't a mention of the native ASQ `visibilityTimeout` feature.

Scooletz:
@SeanFeldman I've rewritten the description to properly represent this approach.

Scooletz:
Could a @Particular/nservicebus-maintainers take a look at this and review it from the `delay delivery` constraint point of view?

SzymonPobiega:
@Scooletz added one comment. Other than that I think it is OK with regards to the core expectations. Can't say much about azure storage part of it though :(

Scooletz:
@SzymonPobiega Thx. I'll take a second look at the pump. Can I distinguish whether or not it's run as a main pump vs satellite ?

That would ASQ target. Opt-in first in minor, then opt-out in major then removal.

SzymonPobiega:
@Scooletz I don't know an easy way to check what pump is it so I just used `settings.LocalAddress` as the input queue in my poller (my poller moves msg from one q to another)

Scooletz:
@SzymonPobiega Could you take a look at f7db403 ? I used the `LocalAddress` to disambiguate whether or not a message pump is run for the main queue (non-satellite).

SzymonPobiega:
@Scooletz looks good, should work.

Scooletz:
I've updated the initial comment to fully describe the implementation.
This provides a full support for `delayed delivery`. IMO - this is huge and needs a in-depth review. I'd like to ship it as opt-in in the next version of ASQ.

src/Tests/Utils.cs
```diff
@@ -0,0 +1,17 @@
...
+    using System;
+
+    public static class Utils
+    {
+        public static string GetEnvConfiguredConnectionString()
+        {
+            return Environment.GetEnvironmentVariable("AzureStorageQueueTransport.ConnectionString");
+        }
+
+        public static string BuildAnotherConnectionString(string connectionString)
```
SeanFeldman: Is it used at all?

src/Transport/Config/AzureStorageTransportExtensions.cs
```diff
@@ -92,6 +92,23 @@ public static TransportExtensions<AzureStorageQueueTransport> DegreeOfReceivePar
...
         }
 
+        /// <summary>
+        /// Switches transport to use timeouts based on the Azure Storage Queues capabilities.
+        /// </summary>
+        public static TransportExtensions<AzureStorageQueueTransport> UseNativeTimeouts(this TransportExtensions<AzureStorageQueueTransport> config, string timeoutTableName)
+        {
+            var settings = config.GetSettings();
+
+            if (string.IsNullOrWhiteSpace(timeoutTableName))
```
SeanFeldman: Table name can be verified using the following regex:  `"^[A-Za-z][A-Za-z0-9]{2,62}$"`

Scooletz:
Ping @SeanFeldman

Scooletz:
@Particular/azure-storage-persistence-maintainers I think that this should go into ` ASQ 7.1`. 
I think it requires a page in doco as well that describes how native delivery works. This PR description provides a good plan how and what to describe.
The nice thing is that this introduced only one method in the public API.

andreasohlund:
This PR must have been closed since I removed the release-6.0.0 branch (since it was merged to master + develop). @Scooletz can you reopen a pull against develop? (I can't seem to be able to re-open + re-target this one)

Sorry!

PR #159, opened 2016-10-10
Add test to show that native integration works

andreasohlund:
This shows how to user the messagewrapper serializer ext point to allow ASQ to receive native json messages.

The problem is that we can't reliably give the message a message id using this strategy.

I'll open a separate pull proposing a new ext point that would expose the full cloud queue message so we can pick the id up from there.

@Particular/azure-storage-queues-maintainers does this make sense?

This PR can be merged as is

### DIFF
--- a/src/AcceptanceTests/Receiving/When_receiving_a_raw_json_message.cs
+++ b/src/AcceptanceTests/Receiving/When_receiving_a_raw_json_message.cs
@@ -1,0 +1,80 @@
+﻿namespace NServiceBus.Azure.Transports.WindowsAzureStorageQueues.AcceptanceTests.Receiving
+{
+    using System;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using Microsoft.WindowsAzure.Storage;
+    using Microsoft.WindowsAzure.Storage.Queue;
+    using Newtonsoft.Json;
+    using NServiceBus.AcceptanceTests;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_receiving_a_raw_json_message : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task ShouldConsumeIt()
+        {
+            var ctx = await Scenario.Define<Context>()
+                .WithEndpoint<Receiver>(b =>
+                {
+
+                    b.When((bus, c) =>
+                    {
+                        var jsonPayload = JsonConvert.SerializeObject(new MyMessage(), Formatting.Indented, new JsonSerializerSettings
+                        {
+                            TypeNameHandling = TypeNameHandling.All
+                        });
+
+                        var message = new CloudQueueMessage(jsonPayload);
+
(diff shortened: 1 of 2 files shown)


andreasohlund:
PR for proposed new extension point raised https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/160

src/AcceptanceTests/Receiving/When_receiving_a_raw_json_message.cs
```diff
@@ -0,0 +1,172 @@
...
+                            TypeNameHandling = TypeNameHandling.All //we need this in order fo $type="x" to be embedded in the json
+                        });
+
+                        var cloudQueueMessage = new CloudQueueMessage(jsonPayload);
+
+                        var connectionString = Environment.GetEnvironmentVariable("AzureStorageQueueTransport.ConnectionString");
+
+                        var storageAccount = CloudStorageAccount.Parse(connectionString);
+                        var queueClient = storageAccount.CreateCloudQueueClient();
+                        var queue = queueClient.GetQueueReference("receivingarawjsonmessage-receiver-retries");
```
SeanFeldman: Why the `retries` queue and not the input queue?

src/AcceptanceTests/Receiving/When_receiving_a_raw_json_message.cs
```diff
@@ -0,0 +1,172 @@
...
+                            SomeProperty = "Test"
+                        };
+                        var jsonPayload = JsonConvert.SerializeObject(message, new JsonSerializerSettings
+                        {
+                            TypeNameHandling = TypeNameHandling.All //we need this in order fo $type="x" to be embedded in the json
+                        });
+
+                        var cloudQueueMessage = new CloudQueueMessage(jsonPayload);
+
+                        var connectionString = Environment.GetEnvironmentVariable("AzureStorageQueueTransport.ConnectionString");
```
SeanFeldman: Use [`Utils.GetEnvConfiguredConnectionString`](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/df33b629903bc92b3f8383cda0f891eb5ecda8a0/src/AcceptanceTests/Utils.cs#L7) instead.

src/AcceptanceTests/Receiving/When_receiving_a_raw_json_message.cs
```diff
@@ -0,0 +1,172 @@
...
+                        stream.Seek(0, SeekOrigin.Begin);
+                        return ReadNative(stream);
+                    }
+                }
+
+                static object[] ReadNative(Stream stream)
+                {
+                    var nativeWrapper = new MessageWrapper();
+
+                    nativeWrapper.Id = Guid.NewGuid().ToString(); //todo: How to deal with this?
```
SeanFeldman: It's a native message, transport message ID is not available, so a random GUID is ok IMO

andreasohlund:
Given that #160 seems to be a better approach I will close this one and clean up that one instead based on @SeanFeldman 's comments

PR #160, opened 2016-10-10
Allow custom envelope unwrapping

andreasohlund:
This PR adds a better extension point to support native integration scenarios where the unwrapper gets full access to the native cloud message.

If we merge this in we should add a native integration sample to docs similar to Sql and Rabbit.

### DIFF
--- a/src/AcceptanceTests/Receiving/When_receiving_a_raw_json_message.cs
+++ b/src/AcceptanceTests/Receiving/When_receiving_a_raw_json_message.cs
@@ -1,6 +1,5 @@
 ﻿namespace NServiceBus.Azure.Transports.WindowsAzureStorageQueues.AcceptanceTests.Receiving
 {
-    using System;
     using System.Collections.Generic;
     using System.IO;
     using System.Threading.Tasks;
@@ -34,11 +33,10 @@
 
                         var cloudQueueMessage = new CloudQueueMessage(jsonPayload);
 
-                        var connectionString = Environment.GetEnvironmentVariable("AzureStorageQueueTransport.ConnectionString");
-
+                        var connectionString = Utils.GetEnvConfiguredConnectionString();
                         var storageAccount = CloudStorageAccount.Parse(connectionString);
                         var queueClient = storageAccount.CreateCloudQueueClient();
-                        var queue = queueClient.GetQueueReference("receivingarawjsonmessage-receiver-retries");
+                        var queue = queueClient.GetQueueReference("receivingarawjsonmessage-receiver");
 
                         return queue.AddMessageAsync(cloudQueueMessage);
                     });
--- a/src/AcceptanceTests/Receiving/When_receiving_a_raw_json_message.cs
+++ b/src/AcceptanceTests/Receiving/When_receiving_a_raw_json_message.cs
@@ -1,0 +1,80 @@
+﻿namespace NServiceBus.Azure.Transports.WindowsAzureStorageQueues.AcceptanceTests.Receiving
+{
+    using System;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using Microsoft.WindowsAzure.Storage;
+    using Microsoft.WindowsAzure.Storage.Queue;
+    using Newtonsoft.Json;
+    using NServiceBus.AcceptanceTests;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_receiving_a_raw_json_message : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task ShouldConsumeIt()
+        {
+            var ctx = await Scenario.Define<Context>()
+                .WithEndpoint<Receiver>(b =>
+                {
+
+                    b.When((bus, c) =>
+                    {
+                        var jsonPayload = JsonConvert.SerializeObject(new MyMessage(), Formatting.Indented, new JsonSerializerSettings
+                        {
+                            TypeNameHandling = TypeNameHandling.All
+                        });
+
+                        var message = new CloudQueueMessage(jsonPayload);
+
+                        var connectionString = Environment.GetEnvironmentVariable("AzureStorageQueueTransport.ConnectionString");
+
+                        var storageAccount = CloudStorageAccount.Parse(connectionString);
+                        var queueClient = storageAccount.CreateCloudQueueClient();
+                        var queue = queueClient.GetQueueReference("receivingarawjsonmessage-receiver-retries");
+
+                        return queue.AddMessageAsync(message);
+                    });
+                })
+                .Done(c => c.GotMessage)
+                .Run();
+
+            Assert.True(ctx.GotMessage);
+        }
+
+        class Context : ScenarioContext
+        {
+            public bool GotMessage { get; set; }
+        }
+
+        class Receiver : EndpointConfigurationBuilder
+        {
+            public Receiver()
+            {
+                EndpointSetup<DefaultServer>();
+            }
+        }
+
+        class MyMessage : IMessage
+        {
+        }
+
+        class MyMessageHandler : IHandleMessages<MyMessage>
+        {
+            public MyMessageHandler(Context ctx)
+            {
+                this.ctx = ctx;
+            }
+
+            public Task Handle(MyMessage message, IMessageHandlerContext context)
+            {
+                ctx.GotMessage = true;
+
+                return Task.FromResult(0);
+            }
+
+            Context ctx;
+        }
+    }
+}
--- a/src/AcceptanceTests/Receiving/When_receiving_a_raw_json_message.cs
+++ b/src/AcceptanceTests/Receiving/When_receiving_a_raw_json_message.cs
@@ -5,7 +5,6 @@
     using System.IO;
     using System.Threading.Tasks;
     using AcceptanceTesting;
-    using AzureStorageQueues;
     using Microsoft.WindowsAzure.Storage;
     using Microsoft.WindowsAzure.Storage.Queue;
     using Newtonsoft.Json;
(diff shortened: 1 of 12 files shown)


src/AcceptanceTests/Receiving/When_receiving_a_raw_json_message.cs
```diff
@@ -0,0 +1,124 @@
...
+                            SomeProperty = "Test"
+                        };
+                        var jsonPayload = JsonConvert.SerializeObject(message, new JsonSerializerSettings
+                        {
+                            TypeNameHandling = TypeNameHandling.All //we need this in order fo $type="x" to be embedded in the json
+                        });
+
+                        var cloudQueueMessage = new CloudQueueMessage(jsonPayload);
+
+                        var connectionString = Environment.GetEnvironmentVariable("AzureStorageQueueTransport.ConnectionString");
```
SeanFeldman: Use `Util.GetEnvConfiguredConnectionString()`

src/AcceptanceTests/Receiving/When_receiving_a_raw_json_message.cs
```diff
@@ -0,0 +1,124 @@
...
+
+                        var storageAccount = CloudStorageAccount.Parse(connectionString);
+                        var queueClient = storageAccount.CreateCloudQueueClient();
+                        var queue = queueClient.GetQueueReference("receivingarawjsonmessage-receiver-retries");
```
SeanFeldman: `-retries` is needed?

src/AcceptanceTests/Receiving/When_receiving_a_raw_json_message.cs
```diff
@@ -0,0 +1,124 @@
...
+                    var wrapper = jsonSerializer.Deserialize<MessageWrapper>(textReader);
+
+                    if (!string.IsNullOrEmpty(wrapper.Id))
+                    {
+                        return wrapper;
+                    }
+
+                    return new MessageWrapper
+                    {
+                        Id = rawMessage.Id,
```
SeanFeldman: 👍 yes, this is much better since natively integrating party could imply with transport message ID the business message ID.

src/Transport/Config/AzureStorageTransportExtensions.cs
```diff
@@ -70,6 +71,15 @@ public static TransportExtensions<AzureStorageQueueTransport> SerializeMessageWr
         }
 
         /// <summary>
+        /// Registers a custom unwrapper to convert native messages to <see cref="MessageWrapper" />. This is needed when receiving raw json/xml/etc messages from non NServiceBus endpoints. 
+        /// </summary>
+        public static TransportExtensions<AzureStorageQueueTransport> UnwrapMessagesWith(this TransportExtensions<AzureStorageQueueTransport> config,IMessageEnvelopeUnwrapper unwrapper)
```
andreasohlund: @Particular/nservicebus-maintainers @SimonCropp didn't really think the extension point through and went with an interface just because that made the internal code easier.

Would a `Func<CloudQueueMessage,MessageWrapper>` be better? (would be one less public type for us?

Other considerations? (thinking we should have some general guidance on lowlevel extensions like this)

Scooletz: I'd say that we can go with a func, as it's highly unlikely that we're going to enhance this translator with other capabilities. The only thing that comes to my mind is that instead of `UnrapMessagesWith` we should have `AddMessageUnwrapper` which would add this func to a collection of functions. If we change the the unwrapping method for all the messages, the endpoint won't be able to deserialize messages sent by other endpoints with regular serializers, will it?

andreasohlund:
>the endpoint won't be able to deserialize messages sent by other endpoints
with regular serializers, will it?

Correct'ish, the one in the test would work for json but not for xml

To have a list we would need a `bool TryUnwrap(CloudQueueMessage, out
MessageWrapper)` right?

Scooletz:
Or just `Func<CloudQueueMessage,MessageWrapper>` returning `null` on failure.

andreasohlund:
I personally think tryget is more intention revealing but I know it doesn't work for async.

Do we forse any async here?

Scooletz:
I think that we agreed that serialization isn't asynchronous in the core, so I can't come up with an idea why it could be here. If you want to `TryGet`, the interface needs to stay though.

danielmarbach:
In general in the core we are leaning towards funcs when possible. I'd say going for additive list sounds overly complex, what would be the benefit?

andreasohlund:
Agreed, a single func seems to be enough for now since you can support all the formats you want in that single func

Scooletz:
I don't agree @danielmarbach I can easily imagine a site that wants to use ASQ with json payloads sent by some clients and integrate with other endpoints using a regular serialization. Why would one need to care to handle both cases, when all they want is to have an option for raw json payloads handled in addition to the regular communication. Just register a higher prio `TryGet` and return false to leave the handling for the transport itself.

danielmarbach:
My point was not the TryGet. users should not need to fiddle with settings. Btw. recoverability override is also not addictive. My point was: Have one func as a simple first shot. When you get real demand from users from real use cases consider adding an additive solution. But only then

Scooletz:
Ok. Let's make it just a `Func<CloudQueueMessage,MessageWrapper>` then.

andreasohlund:
I've now run this for a while and it works nicely.

One thing that doesn't work is that ServiceControl fails to import the audited messages since they don't have a destination:

```
NServiceBus.Azure.Transports.WindowsAzureStorageQueues.EnvelopeDeserializationFailed: Failed to deserialize message envelope ---> System.ArgumentException: Invalid destination address specified
Parameter name: destination
   at NServiceBus.Address.Parse(String destination) in C:\Build\src\NServiceBus.Core\Address.cs:line 91
   at NServiceBus.Azure.Transports.WindowsAzureStorageQueues.AzureMessageQueueReceiver.DeserializeMessage(CloudQueueMessage rawMessage) in c:\BuildAgent\work\4e5353dd260f0a07\src\Transport\AzureMessageQueueReceiver.cs:line 179
   at NServiceBus.Azure.Transports.WindowsAzureStorageQueues.AzureMessageQueueReceiver.Receive() in c:\BuildAgent\work\4e5353dd260f0a07\src\Transport\AzureMessageQueueReceiver.cs:line 110
   --- End of inner exception stack trace ---
   at NServiceBus.Azure.Transports.WindowsAzureStorageQueues.AzureMessageQueueReceiver.Receive() in c:\BuildAgent\work\4e5353dd260f0a07\src\Transport\AzureMessageQueueReceiver.cs:line 114
   at NServiceBus.Azure.Transports.WindowsAzureStorageQueues.PollingDequeueStrategy.Action(Object obj) in c:\BuildAgent\work\4e5353dd260f0a07\src\Transport\PollingDequeueStrategy.cs:line 100

```

Pretty sure that the other transports doesn't enforce this? (cc @Particular/servicecontrol-maintainers )

andreasohlund:
Found the issue with SC. A "integration" message like this will endup up with a null reply to address


```
ï»¿{"IdForCorrelation":null,"Id":"2e7d75ea-cfc5-4e5d-b4bb-c820968028d5","MessageIntent":0,"ReplyToAddress":null,"TimeToBeReceived":"10675199.02:48:05.4775807","Headers":{"NServiceBus.MessageId":"2e7d75ea-cfc5-4e5d-b4bb-c820968028d5","NServiceBus.ContentType":"application/json","NServiceBus.ProcessingStarted":"2016-11-30 07:29:22:758015 Z","NServiceBus.ProcessingEnded":"2016-11-30 07:29:22:904716 Z","$.diagnostics.hostid":"61c5dad898381e7faa58b93e8706cafb","$.diagnostics.hostdisplayname":"RD0003FF729F05","NServiceBus.ProcessingMachine":"RD0003FF729F05","NServiceBus.ProcessingEndpoint":"Backend","NServiceBus.Version":"6.0.0","NServiceBus.TimeSent":"2016-11-30 07:29:22:904716 Z"},"Body":"eyIkdHlwZSI6IkdldEJlZXIuTWF0Y2hpbmcuUmVnaXN0ZXJNYW51YWxNYXRjaCIsIkJlZXJJZCI6IjkwNjYyMCIsIk1hdGNoUmVxdWVzdElkIjoiYm94YmVlci0zODEzMiJ9","CorrelationId":null,"Recoverable":true}

```

and that causes SC to fail the import. Adding a fake reply to address fixed it.

I'll see how the current transport handles this (since SC in on a older version)

andreasohlund:
Workaround to get SC going is to add a fake reply to address header

```
   return new MessageWrapper
                    {
                        Id = rawMessage.Id,
                        Headers = new Dictionary<string, string> {{Headers.ReplyToAddress,"browser"}},//for now since SC can't handle null
                        Body = rawMessage.AsBytes
                    };
```

I've verified that v7 of ASQ can handle this properly so I'd say there is no need to fix this.

src/AcceptanceTests/Receiving/When_receiving_a_raw_json_message.cs
andreasohlund: Fixed

andreasohlund:
API now used a func. 

This one should be good for final review now

src/Transport/IMessageEnvelopeUnwrapper.cs
```diff
@@ -0,0 +1,18 @@
+namespace NServiceBus.AzureStorageQueues
+{
+    using Azure.Transports.WindowsAzureStorageQueues;
+    using Microsoft.WindowsAzure.Storage.Queue;
+
+    /// <summary>
+    /// Extracts a native cloud queue message to the internal message wrapper format.
+    /// </summary>
+    public interface IMessageEnvelopeUnwrapper
```
SeanFeldman: Does it need to be public if user define envelope unwrapper is passed in as a func?

andreasohlund: Good catch, will fix

andreasohlund:
This needs doco and a sample, I'll try to get that going

https://github.com/Particular/docs.particular.net/issues/2389

PR #165, opened 2016-10-18
Remove queue based maximum time of a delayed delivery leaving tables only

Scooletz:
@dvdstelt has raised a concern that for delayed messages that are scheduled for 6 days having a one day to live might be not enough. I lowered the time to `1 day` only leaving 6 days for TTL in a corner case.

@Particular/azure-storage-queues-maintainers As we want to ensure that timeouts are delivered, should we maybe remove usage of a native delay and base it only on the tables?


Scooletz:
Ping @SeanFeldman @yvesgoeleven

dvdstelt:
The idea was that if on a Friday night and endpoint might die

1. When there's little to no monitoring (and this happens A LOT) IT might not figure out it's down until Monday and 2 or 3 days worth of messages might be gone.
2. When there is monitoring, but an endpoint isn't business critical, operations will ignore it until Monday. However very likely without knowing messages will disappear from the queue due to expiration.

We could argue how many days could be appropriate, but I'd rather see how timeouts are being raised. One day might be a good rule, where shorter timeouts get a message in the queue, others get a row/entity in storage. This way we don't need to poll for messages that often, reducing transactions on the storage account. There are customers that will easily spawn thousands of timeouts daily.

Perhaps we should make it an option.
- More reliable, where we always store in table storage
- Less costly (on transactions), where we post messages

SeanFeldman:
Sounds like a good approach. Should be well documented to ensure that customers realize how it works and that leaving timeouts of less than 1d unhandled could result in message loss.

Scooletz:
Let me play the devil's advocate @SeanFeldman  Whether it's 1d or 6d you can have a non-delivered message. ASQ will also drop messages after 7 days on its own (TTL). Do we state it explicitly? If not, there's no difference in this approach.
On the other hand, if there's a doubt related to the `TimeSpan`, maybe we should drop using native deferrals altogether and stick to the tables then?

dvdstelt:
@Scooletz If we can't even support one day, maybe we should drop support for ASQ altogether.... Oh, wait! 😄

Scooletz:
I'd say: stick to the tables then @dvdstelt

SeanFeldman:
Don't need to be a 👿's advocate to bring that point. Scenario that @dvdstelt brought up could be easily extended to a few more days and et voilà. You bring a valid question if we should even bother with native "deferral" or solely rely on the tables. I can't answer that as I don't know what the customers are doing. 

I'm going to draw a comparison between native deferral and system queues. Same problem of 7 days TTL. Difference though, is that customers _can_ use SC/SP or a custom solution to overcome this limitation. With "native deferral" based on a queue that won't be possible.

> On the other hand, if there's a doubt related to the TimeSpan, maybe we should drop using native deferrals altogether and stick to the tables then?

Perhaps we should. A few good question were raised when RabbitMQ maintainers where thinking of how to build native deferral:

1. How at any point in time can a customer know what messages are deferred and for how long?
1. Can they drop those messages if needed?

While the focus was to make native deferral to work, perhaps we are trying to answer the question how those deferred messages are handled from prod-ops point of view.

Scooletz:
You've raised valid points @SeanFeldman Considering all of them, including the prod-ops point of view and visibility of the deferred messages I'd stay, let's remove the timeouts based on ASQ and stick to the tables. We'll get the following:

1. no expiration
1. visibility, if needed (query against a table)
1. simplification of tests and the codebase (dispatch now or store in a table).

Thoughts @Particular/azure-storage-queues-maintainers ?

SzymonPobiega:
@Scooletz @SeanFeldman I am not super sure how important is ability to remove timeouts. If it really is, it rules out probably all native solutions are we are left with a timeout manager again (probably better implemented).

Scooletz:
Not for ASQ, as we use tables from the same account @SzymonPobiega 
This wasn't the only point, still it combines nicely with the rest of points. <oving to `tables-only` seems like a preferable choice whether or not the `visibility argument` stands.

Scooletz:
I changed the title of this issue.

SeanFeldman:
>  I am not super sure how important is ability to remove timeouts. 

@SzymonPobiega not top priority for sure, but a question. The issue I have with ASQ queue-based long time deferral is the built-in 7 days TTL hard limit. It's not going to change. As @Scooletz noted rightfully, it will simplify the code (and potentially troubleshooting). I.e. I don't think having a mix justifies the extra effort.

SzymonPobiega:
@SeanFeldman 👍

dvdstelt:
> How at any point in time can a customer know what messages are deferred and for how long?

That's a really good argument. But when people now select `ASQ` as transport and `SQL` as persister, now they also need to support Azure Tables. Do we think this is an issue, that users are aware of this and whether or not we should provide options? How about a timeout manager in core that solves this? 😝

Scooletz:
They don't `need to support Azure Tables` as with the connection string you can access tables as well. Did you mean, that they'd need to query both? Yes, but you'd query timeouts only in a case of emergency. `TimeoutManager` does not solve it any better as far as I see it.

Scooletz:
Ping @Particular/azure-storage-queues-maintainers I think that we all 👍 this. Could somebody merge it?

dvdstelt:
> Did you mean, that they'd need to query both? Yes, but you'd query timeouts only in a case of emergency. TimeoutManager does not solve it any better as far as I see it.

It means that both **developers** and **operations** team need to have knowledge of the fact that previous **business data** is running in Azure Table Storage, not just in queues and wherever-the-business-data-is-persisted. And that they need to support it. I'm serious, that's no small issue! And we as Particular Software need to support several implementations of this native delivery (ASQ, RabbitMQ, SQL) and also support needs to be made aware of the changes/updates and be aware of the fact that timeouts live there. Which isn't documented anywhere but in a TON of GH issues. That's what I meant, but didn't say. 😃 

I think this completely is the wrong issue (or PR) to discuss this, but if you want advocate of the devil, there you have it 😈 😀
But let's take this offline and merge this.

Scooletz:
Sure @dvdstelt . I've raised an issue for this https://github.com/Particular/NServiceBus.AzureStorageQueues/issues/166 Feel free to chime in and adjust it.

PR #176, opened 2017-01-30
Internalize UseNativeTimeouts 

Scooletz:
This PR makes Native Timeouts internaliz till doco and migration arrives.

### DIFF
--- a/src/AcceptanceTests/Sending/When_delaying_messages_natively.cs
+++ b/src/AcceptanceTests/Sending/When_delaying_messages_natively.cs
@@ -3,6 +3,7 @@
     using System;
     using System.Collections.Generic;
     using System.Diagnostics;
+    using System.Reflection;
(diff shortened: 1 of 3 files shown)


Scooletz:

danielmarbach:
Short question: Have you considered internals visible to attribute for the acceptance tests? What drove you to use reflection? Not saying one is better than the other, just want to understand the reasoning

SeanFeldman:
@danielmarbach we never expose internals to the ATTs project since it's supposed to be treated as if it was production code. I might be off on this.

Scooletz:
Yes @danielmarbach I considered it but had some problems with making it work. This just works and is a quick way of hiding it for now. No need for `InternalVisibleTo` to make all internals visible.

danielmarbach:
Needs a rebase

PR #183, opened 2017-02-03
cache JsonSerializer in TimeoutEntity and add newtonsoft dependency

### LINKED ISSUE #178: TimeoutEntity questions
```
    class TimeoutEntity : TableEntity
    {
        public string Destination { get; set; }
        public byte[] Body { get; set; }
        public string MessageId { get; set; }
        public string Headers { get; set; }
        
        static string Serialize<T>(T obj)
        {
            var sw = new StringWriter();
            new JsonSerializer().Serialize(sw, obj);
            sw.Flush();
            return sw.ToString();
        }

        static T Deserialize<T>(string obj)
        {
            return new JsonSerializer().Deserialize<T>(new JsonTextReader(new StringReader(obj)));
        }
```

 * `new JsonSerializer()` is an expensive operation. and this is on the fast path? So i assume JsonSerializer should be cached?
 * if we explicitly use json .net (ie not an inferred nuget dependency) then it needs to be a nuget dependency in our spec. if we dont, changes in upstream packages in the future could result in no json .net referenced
 * why are we using disposbale stuff and not putting them in a using

SimonCropp:
fixes #178

### DIFF
--- a/packaging/nuget/nservicebus.azurestoragequeues.nuspec
+++ b/packaging/nuget/nservicebus.azurestoragequeues.nuspec
@@ -16,6 +16,7 @@
     <dependencies>
       <dependency id="WindowsAzure.Storage" version="[8.0.0, 9.0.0)" />
       <dependency id="NServiceBus" version="[6.0.0, 7.0.0)" />
+      <dependency id="Newtonsoft.Json" version="[9.0.0, 10.0.0)" />
     </dependencies>
   </metadata>
   <files>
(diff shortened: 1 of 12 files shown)


SimonCropp:
@Particular/azure-maintainers thoughts?

SimonCropp:
But pr means nothing after a merge. Only the commits matter.

danielmarbach:
See https://github.com/Particular/NServiceBus.Persistence.AzureStorage/pull/127

Scooletz:
@danielmarbach it's marked as `Accepted` so I'm leaving it for you to merge.

Issue #181, opened 2017-02-02
v5 endpoints cannot receive messages sent by v6 SendOnly endpoints


mikeminutillo:
## Who's affected
All users sending messages from V6 SendOnly endpoints to V5 endpoints.

## Symptoms
Message sent from V6 SendOnly endpoint will cause exception when processed by receiving V5 endpoint and ultimately result in message being moved to the `error` queue.

## Details 
This problem is caused by the fact that V6 SendOnly endpoint does not add `ReplyToAddress` to it's outgoing message headers and the V5 receiver is expecting a non-null ReplyToAddress (see [here](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/support-6.2/src/Transport/AzureMessageQueueReceiver.cs#L184)).

NOTE: A V5 SendOnly endpoint *is* able to send messages to a V5 receiver. This is because the ReplyToAddress gets filled in [here](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/master-6.2.1-bump-up/src/Transport/AzureMessageQueueSender.cs#L111)

_Edit by @boblangley_
Raised from Customer case 00023775

mikeminutillo:
Repro is https://github.com/mikeminutillo/Asq181Repro

Note https://github.com/mikeminutillo/Asq181Repro/blob/master/SenderV6/Program.cs#L22

Scooletz:
Investigation:

I followed this issue and checked how the `ReplyTo` header was appended in v5 Core. It followed the same principles as in v6 with not appending this header for `SendOnly` endpoints (it was done not by a pipeline behavior, still resulting in the same outcome). Next, I checked this header is applied for ASQ in v6 and there it's [done conditionally](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/master/src/Transport/Dispatcher.cs#L119)

```
var toSend = new MessageWrapper
{
    //...
    ReplyToAddress = headers.GetValueOrDefault(Headers.ReplyToAddress),
    //...
};
```

which is different from the `v5` that was [done unconditionally](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/support-6.2/src/Transport/AzureMessageQueueSender.cs#L111):

```
var replyToAddress = validation.Determine(config.Settings, message.ReplyToAddress ??
        options.ReplyToAddress ?? config.LocalAddress, config.TransportConnectionString());
```

[...]

PR #184, opened 2017-02-03
Hotfix 6.2.2 - accept messages with empty ReplyTo

Scooletz:
This PR enables processing messages that have no `ReplyTo` value stored in the `MessageWrapper` structure. `Send-only` endpoints with  `ASQ 7.*` and higher will store messages with no `ReplyTo` causing `ASQ 6.*` to treat these messages as poisonous. 

This PR solves #181


SeanFeldman:
Is this a WIP? Based on the commits I don't see anything changed in the transport...

Also, removing items from ATT FW will only help till the next update when they are all back.

Scooletz:
This is still a WIP. I updated the title.
Removing them ensured a progress. Now I have a green build that I can move forward with. I'm happy to discuss how to make this old branch green without removals. After discussing it with @danielmarbach I'm seeing not that much value in spending more time on ATT that it's needed to make it green though.

Scooletz:

src/AcceptanceTests/Sending/When_message_is_sent_with_no_reply_to_header.cs
```diff
@@ -0,0 +1,113 @@
...
+    public class When_message_is_sent_with_no_reply_to_header : NServiceBusAcceptanceTest
+    {
+        CloudQueue helperQueue;
+        CloudQueue testQueue;
+        string connectionString;
+
+        [SetUp]
+        public void SetUp()
+        {
+            connectionString = Environment.GetEnvironmentVariable("AzureStorageQueueTransport.ConnectionString");
```
danielmarbach: Should we get it from user scope?

src/AcceptanceTests/Sending/When_message_is_sent_with_no_reply_to_header.cs
```diff
@@ -0,0 +1,113 @@
...
+            testQueue.CreateIfNotExists();
+        }
+
+        [Test]
+        public void Should_dispatch_properly()
+        {
+            var context = new Context();
+
+            Scenario.Define(context)
+                .WithEndpoint<ReceiverEndPoint>(b =>
```
danielmarbach: I'd just call this receiver

src/Transport/AzureMessageQueueReceiver.cs
```diff
@@ -181,14 +181,25 @@ TransportMessage DeserializeMessage(CloudQueueMessage rawMessage)
...
+                }
+                else
                 {
-                    Body = m.Body ?? new byte[0],
-                    CorrelationId = m.CorrelationId,
-                    Recoverable = m.Recoverable,
-                    TimeToBeReceived = m.TimeToBeReceived,
-                    MessageIntent = m.MessageIntent
-                };
+                    transportMessage = new TransportMessage(m.Id, m.Headers, Address.Parse(m.ReplyToAddress));
```
danielmarbach: This seems to be using an already obsoleted with warn member https://github.com/Particular/NServiceBus/blob/support-5.2/src/NServiceBus.Core/TransportMessage.cs#L54

danielmarbach: It feels strange to do this if statement here since the core already has it when you access the ReplyToAddress.

https://github.com/Particular/NServiceBus/blob/support-5.2/src/NServiceBus.Core/TransportMessage.cs#L108

I might be misunderstanding something here but according to my understanding when the headers contain a reply to address it will be automatically parsed by the transport message when you access `ReplyToAddress` property.

src/AcceptanceTests/Sending/When_message_is_sent_with_no_reply_to_header.cs
SeanFeldman: Since there's [`Utils.GetEnvConfiguredConnectionString()`](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/926e177f1d60be5d172d3f80d3c889817dfe66b3/src/AcceptanceTests/Utils.cs#L7) it should probably be used and scope to `USER`.

src/AcceptanceTests/Sending/When_message_is_sent_with_no_reply_to_header.cs
```diff
@@ -0,0 +1,113 @@
...
+        string connectionString;
+
+        [SetUp]
+        public void SetUp()
+        {
+            connectionString = Environment.GetEnvironmentVariable("AzureStorageQueueTransport.ConnectionString");
+            var account = CloudStorageAccount.Parse(connectionString);
+            var queues = account.CreateCloudQueueClient();
+            helperQueue = queues.GetQueueReference("custom-no-reply-header");
+            testQueue = queues.GetQueueReference("messageissentwithnoreplytoheader-receiverendpoint");
```
SeanFeldman: Using core Conventions class to get an endpoint name would save from hussle when class is renamed.

src/Transport/AzureMessageQueueReceiver.cs
SeanFeldman: @danielmarbach to confirm what you're highlighting, just constructing `TransportMessage` and then `reply-to` header set conditionally on it should achieve the same.

danielmarbach:
@SeanFeldman wouldn't that mean we could also just use a mutator which removes the header and then use a normal acceptance test with a send local to simplify the current complex setup for the ATT?

SeanFeldman:
@danielmarbach in the context of ATT, yes, agree. I did not think of it that way. It will make ATT setup much cleaner/simpler.

SeanFeldman:
Have opened PR #186 against this PR to address @danielmarbach 's comment.

danielmarbach:
I made a terrible job in explaining myself. I think the fix should be just:

`return new TransportMessage(m.Id, m.Headers)`

instead of 

`return new TransportMessage(m.Id, m.Headers, Address.Parse(m.ReplyToAddress))`

Remove the if else stuff. The condition is already done by the core. And in that case I would even argue it doesn't need a test.

src/AcceptanceTests/Sending/When_message_is_sent_with_no_reply_to_header.cs
Scooletz: The `Utils` you mentioned @SeanFeldman are in `v7` and hasn't been ported to this branch. I'll use the user scope though.

Scooletz: This util is in `develop` (`7.0`) not in `6.2`, but I'll add the user scoping.

Scooletz:
> The condition is already done by the core. And in that case I would even argue it doesn't need a test.

I don't agree with this statement @danielmarbach This test checks "what if MessageWrapper has `ReplyTo` set to null" not "what if the `TransportMessage` header `ReplyTo` is set to null". It's purely ASQ specific. Having this said and being able to use this test for reproducing and then fixing the bug (red->green build), I'm leaving it with some minor changes applied after your comments.

danielmarbach:
@Scooletz ok I missed the fact that we have [ReplyToAddress on the wrapper](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/develop/src/Transport/MessageWrapper.cs#L15). Thanks for mentioning that. I'll withdraw what I previously said

Scooletz:
Ping @danielmarbach could you review it again and apply remarks/change the review outcome?
Ping @Particular/azure-maintainers  The build is still ongoing, but would like to get your reviews :)

src/Transport/AzureMessageQueueReceiver.cs
```diff
@@ -181,24 +181,24 @@ TransportMessage DeserializeMessage(CloudQueueMessage rawMessage)
                     throw new SerializationException("Failed to deserialize message with id: " + rawMessage.Id);
                 }
 
-                TransportMessage transportMessage;
-
                 if (string.IsNullOrEmpty(m.ReplyToAddress))
                 {
-                    transportMessage = new TransportMessage(m.Id, m.Headers);
+                    m.Headers.Remove(Headers.ReplyToAddress);
```
danielmarbach: Could we end up in a scenario where the wrapper does not contain the reply address but the header contains it and then we strip it away?

tmasternak:
The changes to the LGTM. The changes to getting connectionString from `USER` environment variables appear not to work on [Build Server](https://builds.particular.net/viewLog.html?buildId=206168&buildTypeId=NServiceBus_Transports_NServiceBusAzureStorageQueues_1Build&tab=buildLog#_focus=1846)

src/Transport/AzureMessageQueueReceiver.cs
Scooletz: Yes, that would be the case. Let me check if this duality is possible (it should not).

Scooletz:
@tmasternak Let me address the build and the clarification for https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/184#discussion_r100002903 in one go

Scooletz:
I'm posting this as a PR comment as it included important reasoning for the case brought by @danielmarbach in https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/184#discussion_r100002903

> Could we end up in a scenario where the wrapper does not contain the reply address but the header contains it and then we strip it away?

In `ASQ v6` (core v5), the reply header is always added to the headers and the message wrappers. So there's no duality in here nor possibility of having the `MessageWrapper` having `null` in `ReplyTo`.

In `ASQ v7` (core v6), the reply header for `MessageWrapper` is assigned using [the following line](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/develop/src/Transport/Dispatcher.cs#L134):

```
ReplyToAddress = headers.GetValueOrDefault(Headers.ReplyToAddress),
```
Which sets it only when headers contains a value for this header. Effectively it's either:
- the header's non-null value set in both, the wrapper and the message's headers
- the header's value set to `null` in both, the wrapper and the message's headers.

Does it addresses your concern @danielmarbach ?

danielmarbach:
@Scooletz good for me

Scooletz:
@SeanFeldman Is there anything in your comment that has not been dressed?

PR #186, opened 2017-02-07
Updated test to use mutator

SeanFeldman:
Testing using mutator.

Szymon, I've noticed that when I don't remove the header (i.e. remove mutator) the test is still green. Something is off, but too tired to chase it now.


Scooletz:
None of `mutators` can be applied to reproduce the missing `ReplyTo`. The level in which it is stored is the `MessageWrapper` class that is serialized by the ASQ Transport to/from the message queue. There's no point in the core nor in ASQ Transport itself where you can inject a mutator or anything else and change the `MessageWrapper`:
- before serializing and adding it to the queue
- before turning it into `TransportMessage` and passing to the pipeline.

I'm closing this PR.

PR #190, opened 2017-02-09
[WIP] Fix for scale-out scenario with no account aliases enabled

Scooletz:
## Who's affected

Anyone using Azure Storages Queue transport and using multiple storage accounts without specifying storage account aliases.

## Symptoms

`System.Collections.Generic.KeyNotFoundException` exception is raised with the the following error: 

"No account was mapped under following name 'DefaultEndpointsProtocol=https;...'. Please map it using
AddStorageAccount method."

## Original issue

This PR fixes the case where multiple storage accounts are used for scale-out scenario with no `account aliases` enabled.

This bug causes no message loss. The address mapping occurs before the `ReceiveStrategy` kicks in, meaning, that even for `AtMostOnce` we don't loose messages.


Scooletz:
Closed in favor of #193

PR #193, opened 2017-02-10
Fix for scale-out scenario with no account aliases enabled

Scooletz:
## Who's affected

Anyone using Azure Storages Queue transport and using multiple storage accounts without specifying storage account aliases.

## Symptoms

`System.Collections.Generic.KeyNotFoundException` exception is raised with the the following error: 

"No account was mapped under following name 'DefaultEndpointsProtocol=https;...'. Please map it using
AddStorageAccount method."

## Original issue

This PR fixes the case where multiple storage accounts are used for scale-out scenario with no `account aliases` enabled.

This bug causes no message loss. The address mapping occurs before the `ReceiveStrategy` kicks in, meaning, that even for `AtMostOnce` we don't loose messages.


Scooletz:
This PR has been reopened because of the wrongly assigned branch version.

Scooletz:
The PR consists of 4 commits that are supposed to be named and described properly. 2 first bumps up dependencies, 3rd introduced a failing test, 4th - fixes it.

src/AcceptanceTests/Configuration/When_receiving_from_another_account_wo_aliases.cs
```diff
@@ -0,0 +1,66 @@
+﻿namespace NServiceBus.AcceptanceTests.WindowsAzureStorageQueues.Configuration
+{
+    using System;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_receiving_from_another_account_wo_aliases : NServiceBusAcceptanceTest
```
SeanFeldman: While from a memory optimization "wo" is less bytes than "without", I'd still go with the expensive version 😉

src/AcceptanceTests/Configuration/When_receiving_from_another_account_wo_aliases.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        const string ReplyToValue = "queue@DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;";
```
SeanFeldman: `ReplyToValue` is `UnmappedConnectionString`?

src/AcceptanceTests/Configuration/When_receiving_from_another_account_wo_aliases.cs
Scooletz: Nice one:P Unfortunately, it's not memory but too long path on our `TC` agent. I needed to trim it to make it work.

Scooletz:

PR #161, opened 2016-10-11
Delayed delivery native

Scooletz:
Related to https://github.com/Particular/NServiceBus/issues/4085
Solves the following natively in ASQ: https://github.com/Particular/NServiceBus.Persistence.AzureStorage/issues/6
Elects single timeout poller when scaled out: https://github.com/Particular/NServiceBus/issues/2602

This PR provides a native support for `delayed delivery` in ASQ.

Summary:
- when a timeout delay is lower than `MaxVisibilityDelay`, a native ASQ mechanism `visibilityTimeout` is used
- when a timeout delay is greater than `MaxVisibilityDelay`, a `TimeoutEntity` is written to a dedicated table in a format that enables to recreate a message to be sent
-  `MaxVisibilityDelay = CloudQueueMessage.MaxTimeToLive - TimeSpan.FromDays(1)` - just to not expire timeouts just before dispatching
- a custom `TimeoutsPoller` is provided to query the store and dispatch messages to the target queues
- if a target queue does not exists (for any reason, like an endpoint disposal), the poller will place the message in the endpoints` queue
- a blob lease lock manager is implemented to run a single `TimeoutsPoller` for every endpoint, across its all instances ([original ASP issue](https://github.com/Particular/NServiceBus.Persistence.AzureStorage/issues/6))

Abandoned ideas:
- the initial idea of using `AppendOnlyBlob` was abandoned as it enables only 50,000 blocks. Creation of new blobs for a specific time period etc would be much more harder than managing a table + a blob based lock.

_This is a continuation of https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/155_

PR #198, opened 2017-02-24
[WIP]Spike native pub sub

andreasohlund:
Connects to #277 

Switching the transport to native pubsub storing subscription details in table storage.

This is just a hack to prove that it works

### DIFF
--- a/src/AcceptanceTests/ConfigureAzureStorageQueueTransport.cs
+++ b/src/AcceptanceTests/ConfigureAzureStorageQueueTransport.cs
@@ -25,21 +25,12 @@ public class ConfigureEndpointAzureStorageQueueTransport : IConfigureEndpointTes
     public Task Configure(string endpointName, EndpointConfiguration configuration, RunSettings settings, PublisherMetadata publisherMetadata)
     {
         var connectionString = settings.Get<string>("Transport.ConnectionString");
-        var transportConfig = configuration
+            var transportConfig = configuration
             .UseTransport<AzureStorageQueueTransport>()
             .ConnectionString(connectionString)
             .MessageInvisibleTime(TimeSpan.FromSeconds(30));
-        //.SerializeMessageWrapperWith<JsonSerializer>();
 
-        var routingConfig = transportConfig.Routing();
 
-        foreach (var publisher in publisherMetadata.Publishers)
-        {
-            foreach (var eventType in publisher.Events)
-            {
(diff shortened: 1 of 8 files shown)


andreasohlund:
The 4 failing tests https://builds.particular.net/viewLog.html?buildId=208232&buildTypeId=NServiceBus_Transports_NServiceBusAzureStorageQueues_1Build is related to multi inheritance but that should be easy to fix

andreasohlund:
@Particular/azure-maintainers my idea was a mixed mode option so that we can keep the current message driven pubsub for backwards compatibility but new users could use this new simpler approach?

SeanFeldman:
@andreasohlund like the idea.

andreasohlund:
@SzymonPobiega @timbussmann ideas on how to support mixed mode, how to deal with [`IMessageDrivenSubscriptionTransport`](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/IMessageDrivenSubscriptionTransport.cs)

Scooletz:
We're moving more and more to this storage account of ASQ. Good 👍

andreasohlund:
I didn't know the native timeouts where already in https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/161 but since that has happened this PR would mean ASQ no longer have a hard dep on ASP to work properly?

Scooletz:
They were, but they are not in public API as we need to provide:
- doco
- migration
- review the code again.

Once we have it, yes, beside sagas, we need no storage.

andreasohlund:
Do we want to proceed with this? (if yes that I'll get the caching and config options in)

SeanFeldman:
It's marked as in-progress, so I've assigned it to you (since you're the originator of the PR).

> Do we want to proceed with this?
Depends who you ask 😉

SeanFeldman:
@boblangley will raise an issue in PlatDev for this (pub/sub) and native timeouts.

SeanFeldman:
@andreasohlund was this spike supposed to go into transport v8?

andreasohlund:
Was just trying to prove that it was a low hanging fruit. I guess it can be added in any minor version but deprecating support for message driven pub sub would have to wait.

In short: This can wait until the netcore version is released IMO

SeanFeldman:
Closing this for now. Please reopen when and if it becomes relevant again.

PR #213, opened 2017-05-24
Proper check for nullable cancellation token source

Scooletz:
When `nativeTimeouts` are disabled, `Stop` threw an exception as the cancellation source. This PR fixes it.


Scooletz:

PR #216, opened 2017-05-31
Replacing time consuming and not deterministic cleaning queues with testId aware approach

Scooletz:
This PR removes the previous solution for cleaning queues between tests based on issuing clean command to all the queues in the account and replaces it with a custom `testId` header that is added to every message sent to any queue. On the incoming side, all the messages not from the current run are filtered out.

It's done by augmenting the `MessageWrapper`. This unfortunately required some work related to tests that test the message wrappers themselves. Additionally, a few tests for aliases (aka _secure connection strings_) were modified as previously they were based on `Peek`ing a message, which now, when queues are not cleared, is no longer suitable. It was replaced by a loop spinning till either a message from the current run is got or the audit queue is empty.

#### Another solutions that were considered but eventually were dropped

Adding outgoing behavior augmenting `IDispatchContext` was tried. This failed terribly as `IDispatchMessages` is used not only by the dispatching pipeline but also separate pipelines for subscribing and unsubscribing for message-driven pubsub.

Introducing possibility of adding additional headers to the dispatcher was postponed as it created an additional extension point not needed for anything else than passing this test identifier.

### DIFF
--- a/src/AcceptanceTests/Configuration/When_sending_messages_with_mapped_account_names.cs
+++ b/src/AcceptanceTests/Configuration/When_sending_messages_with_mapped_account_names.cs
@@ -88,15 +88,33 @@
 
             Assert.IsTrue(ctx.Received);
 
-            var rawMessage = await auditQueue.PeekMessageAsync();
-            JToken message;
-            using (var reader = new JsonTextReader(new StreamReader(new MemoryStream(rawMessage.AsBytes))))
+            Dictionary<string, string> propertiesFlattened;
+            do
             {
-                message = JToken.ReadFrom(reader);
-            }
+                var rawMessage = await auditQueue.GetMessageAsync();
+                await auditQueue.DeleteMessageAsync(rawMessage);
+                if (rawMessage == null)
+                {
+                    Assert.Fail("No message in the audit queue to pick up.");
+                }
 
-            ctx.AllPropertiesFlattened = message.FindProperties(IsSimpleProperty)
-                .ToDictionary(jp => jp.Name, jp => ((JValue) jp.Value).Value<string>());
+                JToken message;
+                using (var reader = new JsonTextReader(new StreamReader(new MemoryStream(rawMessage.AsBytes))))
+                {
+                    message = JToken.ReadFrom(reader);
+                }
+
+                propertiesFlattened = message.FindProperties(IsSimpleProperty)
+                    .ToDictionary(jp => jp.Name, jp => ((JValue)jp.Value).Value<string>());
+
+                if (propertiesFlattened.ContainsValue(ctx.TestRunId.ToString()))
+                {
+                    break;
+                }
+            } while (true);
+
+
+            ctx.AllPropertiesFlattened = propertiesFlattened;
 
             ctx.ContainingRawConnectionString = ctx.AllPropertiesFlattened.Where(kvp => kvp.Value.Contains(defaultConnectionString))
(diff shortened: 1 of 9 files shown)


src/AcceptanceTests/Receiving/When_receiving_a_raw_json_message.cs
```diff
@@ -85,15 +85,17 @@ static MessageWrapper MyCustomUnwrapper(CloudQueueMessage rawMessage)
                 return new MessageWrapper
                 {
                     Id = rawMessage.Id,
-                    Headers = new Dictionary<string, string>(),
+                    Headers = new Dictionary<string, string>
+                    {
+                        {TestIndependence.HeaderName,contextTestRunId.ToString() }
```
danielmarbach: Formatting of brackets is off

src/AcceptanceTests/Configuration/When_sending_messages_with_mapped_account_names.cs
```diff
@@ -88,15 +88,33 @@ public When_sending_messages_with_mapped_account_names()
...
+                }
+
+                propertiesFlattened = message.FindProperties(IsSimpleProperty)
+                    .ToDictionary(jp => jp.Name, jp => ((JValue)jp.Value).Value<string>());
+
+                if (propertiesFlattened.ContainsValue(ctx.TestRunId.ToString()))
+                {
+                    break;
+                }
+            } while (true);
```
danielmarbach: So in theory this might idle forever? Should we have a timeout condition here?

src/Transport/Config/AzureStorageTransportExtensions.cs
```diff
@@ -138,6 +139,12 @@ internal static TransportExtensions<AzureStorageQueueTransport> UseNativeTimeout
             return config;
         }
 
+        internal static TransportExtensions<AzureStorageQueueTransport> UseAddtionalHeaders(this TransportExtensions<AzureStorageQueueTransport> config, Lazy<Dictionary<string, string>> additionalHeaders)
```
danielmarbach: Is this change relevant? I couldn't find where it is used

src/AcceptanceTests/Configuration/When_sending_messages_with_mapped_account_names.cs
Scooletz: Nope it will fail as soon as the `auditQueue` is drained. See at line 98.

src/AcceptanceTests/Configuration/When_sending_messages_with_mapped_account_names.cs
```diff
@@ -88,15 +88,33 @@ public When_sending_messages_with_mapped_account_names()
...
+            Dictionary<string, string> propertiesFlattened;
+            do
             {
-                message = JToken.ReadFrom(reader);
-            }
+                var rawMessage = await auditQueue.GetMessageAsync();
+                await auditQueue.DeleteMessageAsync(rawMessage);
+                if (rawMessage == null)
+                {
+                    Assert.Fail("No message in the audit queue to pick up.");
```
Scooletz: Stop condition for this loop @danielmarbach

src/Transport/Config/AzureStorageTransportExtensions.cs
Scooletz: Will remove.

Scooletz:
Ping @danielmarbach

PR #219, opened 2017-06-07
update to new VS 2017 csproj format

SimonCropp:
this is in prep for targeting net standard.

PR #220, opened 2017-06-07
Native Delayed Delivery support for ASQ

Scooletz:
Connected with https://github.com/Particular/PlatformDevelopment/issues/852

This is a major release of **ASQ 8.0**

This PR introduces Native Delayed Delivery based on the Azure Table Storage. It changes the way timeouts are handled:
- by default, all future timeouts will be delivered via Native Delayed Delivery. No new timeouts are scheduled via persister's timeout API
- `DelayedDeliverySettings.DisableTimeoutManager` should be called when all the timeouts registered with persister's timeout API were dispatched. This settings disables running the core's timeouts poller.


Scooletz:
@Particular/azure-maintainers I'd like to ask you for an opinion. 

As this is now a part of moving all the timeouts to natively supported delayed delivery when possible, I followed the same approach as in others and proposed the API for passing a suffix that would be added to the queue name, to make the table unique per endpoint. I've set the suffix to `timeouts` for all the test by using the [new configuration](https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/220/files#diff-b3f64d093e4b15bb575d0d519ac4b5e9R25). 
As queue names can contain `'-'` I tried to apply some encoding but then, I was hit by some tests having their queues longer than 63 characters, which are shortened, but then again, after appending `timeouts` suffix were longer than the limit of characters per table.
What would be your proposal to fix it?

Scooletz:
I've changed the approach to explicitly stating the timeout table name.

Scooletz:

danielmarbach:
Needs a rebase because of #219

src/AcceptanceTests/ConfigureAzureStorageQueueTransport.cs
```diff
@@ -19,7 +19,11 @@ public Task Configure(string endpointName, EndpointConfiguration configuration,
             .UseTransport<AzureStorageQueueTransport>()
             .ConnectionString(connectionString)
             .MessageInvisibleTime(TimeSpan.FromSeconds(30));
-        
+
+        var delayedDelivery = transportConfig.DelayedDelivery();
+        delayedDelivery.DisableTimeoutManager();
```
danielmarbach: So the disable timeouts explicit call is needed for backward compatibility/ I was wondering if defining TableSuffix would be enough?

src/Transport/AzureStorageQueueInfrastructure.cs
```diff
@@ -22,30 +22,26 @@ internal AzureStorageQueueInfrastructure(ReadOnlySettings settings, string conne
...
+        public override IEnumerable<Type> DeliveryConstraints
+        {
+            get
             {
-                var @true = Task.FromResult(true);
-                shouldDispatch = (u,y) => @true;
+                yield return typeof(DiscardIfNotReceivedBefore);
+                yield return typeof(NonDurableDelivery);
+
+                if (delayedDeliverySettings.TimeoutManagerDisabled)
```
danielmarbach: if a user doesn't call `DisableTimeoutManager()` then those two delivery constraints are not supported correct? Can you elaborate the design decision?

src/Transport/AzureStorageQueueInfrastructure.cs
```diff
@@ -181,14 +174,22 @@ public override Task Stop()
             return poller != null ? poller.Stop() : TaskEx.CompletedTask;
         }
 
+        string GetDelayedQueueTableName()
+        {
+            if (string.IsNullOrEmpty(delayedDeliverySettings.Suffix))
+            {
+                throw new Exception("Native delayed delivery feature requires configuring a table suffix.");
```
danielmarbach: If we always require a table suffix why don't we make it mandatory in the `DelayedDeliver `extension method as a parameter?

src/Transport/Config/AzureStorageTransportExtensions.cs
```diff
@@ -115,27 +114,13 @@ public static TransportExtensions<AzureStorageQueueTransport> DegreeOfReceivePar
...
 
         /// <summary>
-        /// Switches transport to use timeouts based on the Azure Storage Queues capabilities.
+        /// 
         /// </summary>
-        // TODO: UseNativeTimeouts are internalized till doco and migration arrives.
-        internal static TransportExtensions<AzureStorageQueueTransport> UseNativeTimeouts(this TransportExtensions<AzureStorageQueueTransport> config, string timeoutTableName)
+        /// <param name="config"></param>
+        /// <returns></returns>
+        public static DelayedDeliverySettings DelayedDelivery(this TransportExtensions<AzureStorageQueueTransport> config)
```
danielmarbach: Make the suffix mandatory here?

src/Transport/Config/DelayedDeliverySettings.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// <summary>
+        /// Sets the suffix for the table storing delayed messages.
+        /// </summary>
+        /// <param name="suffix"></param>
+        public void TableSuffix(string suffix)
+        {
+            Guard.AgainstNullAndEmpty(nameof(suffix), suffix);
+
+            const string tableNameRegex = "^[A-Za-z][A-Za-z0-9]{2,62}$";
+            if (new Regex(tableNameRegex).IsMatch(suffix) == false)
```
danielmarbach: Not that it really matters here but how about making the regex static in this class and use the static regex?

src/Transport/Config/DelayedDeliverySettings.cs
```diff
@@ -0,0 +1,39 @@
...
+            {
+                throw new ArgumentException($"{nameof(suffix)} must match the following regular expression '{tableNameRegex}'");
+            }
+            
+            Suffix = suffix;
+        }
+
+        /// <summary>
+        /// Disables the Timeout Manager for the endpoint. Before disabling ensure there all timeouts in the timeout store have been processed or migrated.
```
danielmarbach: Ok I think this answers the question I raised at the beginning about the disable call

src/Transport/PreventRoutingMessagesToTimeoutManager.cs
```diff
@@ -0,0 +1,17 @@
...
+    class PreventRoutingMessagesToTimeoutManager : Feature
+    {
+        public PreventRoutingMessagesToTimeoutManager()
+        {
+            EnableByDefault();
+        }
+
+        protected override void Setup(FeatureConfigurationContext context)
+        {
+            context.Pipeline.Remove("RouteDeferredMessageToTimeoutManager");
```
danielmarbach: For my education: What are the consequences if this feature is enabled and the timeout manager is also enabled? Does it mean all new timeouts will use the native way but existing timeouts in the TM will still be polled?

danielmarbach: Since we are moving to native timeouts more and more should this be something that the core provides without us needing to remove core behaviors explicitly? @Particular/nservicebus-maintainers

danielmarbach:
@Scooletz made a few comments. Some might have been collapsed by GH. Don't know why

src/AcceptanceTests/ConfigureAzureStorageQueueTransport.cs
Scooletz: It's a full table name right now as elaborated in https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/220#issuecomment-306772203 Making this explicitly set, makes user responsible for new timeouts. I see no reason why that would be not enough? Could you elaborate?

src/Transport/AzureStorageQueueInfrastructure.cs
Scooletz: Sure. When you don't disable `TimeoutManager`, these two delivery constraints are not added, therefore the `TimeoutManager` will be enabled and dispatching timeouts registered in the old-fashioned way (a link to the core code https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/DelayedDelivery/DelayedDeliveryFeature.cs#L24 ).

New timeouts will be registered in the new way because we remove the timeout related behavior that was sending it to the satellite and storing it in the persistence.

src/Transport/PreventRoutingMessagesToTimeoutManager.cs
andreasohlund: >Since we are moving to native timeouts more and more should this be something that the core provides without us needing to remove core behaviors explicitly?

This has already been mentioned by @bording , perhaps this can be in scope for v7 if the effort is reasonable?

src/Transport/AzureStorageQueueInfrastructure.cs
Scooletz: I wanted to keep this API aligned with SQL PR, but I might change it as it looks reasonable. Could you take a look at this as well? https://github.com/Particular/NServiceBus.SqlServer/pull/364/files#diff-737538bdeba63b408be4741aec8c4ae3R33

src/Transport/Config/DelayedDeliverySettings.cs
Scooletz: It doesn't really matter (usually called once at startup)/

src/Transport/PreventRoutingMessagesToTimeoutManager.cs
Scooletz: You are very correct sir in your first comment.

Daniel, we had this discussion with TF under https://github.com/Particular/PlatformDevelopment/issues/852 . Our conclusion was that it will take some time to make this happen in the core, but still, whenever a possibility like this arrives, we can easily integrate it in `Native Delayed Delivery` implementations when needed.

Scooletz:
@danielmarbach Thank you for pointing out github issue :) I replied to all of them.

src/Transport/Config/DelayedDeliverySettings.cs
danielmarbach: I'm good with it, see my intro:

> Not that it really matters

src/Transport/AzureStorageQueueInfrastructure.cs
danielmarbach: Made a comment over there as well

src/AcceptanceTests/ConfigureAzureStorageQueueTransport.cs
danielmarbach: I was wondering when I made the comment if calling `DelayedDelivery` would automatically disable the timeout manager. Later I realized we can't because of backward compat.

Scooletz:
This is rebased and ready for a review

bording:
@Scooletz Spent some time looking at this, and I think there are several things that are missing:

- I see that you're disabling the timeout manger behavior, which gets you saga timeouts and deferred messages sent natively, but I don't see any place where you're checking for timeout manager headers when calculating your delay value. This means that delayed retries will still be sent to the timeout manager, and will then break completely when someone calls `DisableTimeoutManager`. See https://github.com/Particular/NServiceBus.RabbitMQ/blob/develop/src/NServiceBus.Transport.RabbitMQ/Sending/BasicPropertiesExtensions.cs#L108 for an example.

- The ordering of your delay calculation should be swapped to be consistent with previous TM behavior. Core doesn't not currently prevent a message from having both delivery constraints, and the TM checks for `DoNotDeliverBefore` first, then `DelayDeliveryWith`: 
https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/DelayedDelivery/RouteDeferredMessageToTimeoutManagerBehavior.cs#L58

- The TM doesn't allow TTBR to be set on a message that has a delay, so I suggest respecting that as well and throwing an exception instead: https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/DelayedDelivery/RouteDeferredMessageToTimeoutManagerBehavior.cs#L32

- There are several TM-related settings that can have some weird interactions when used in conjunction with this native delivery stuff because they all impact if the TM is active or not:
  - Send-only endpoints
  - external timeout manager address
  - Explicitly disabling the feature via `DisableFeature`

The external address being configured leads to some REALLY weird behavior, so it should never be allowed. The other two end up with the TM disabled but the transport hasn't told core it supports native delivery (via DeliveryConstraints) so things end up in a rather broken state.

Take a look at https://github.com/Particular/NServiceBus.RabbitMQ/blob/develop/src/NServiceBus.Transport.RabbitMQ/DelayedDelivery/DelayInfrastructure.cs#L92 to see what I'm talking about. That's done as a startup check.

The other part of that is here: https://github.com/Particular/NServiceBus.RabbitMQ/blob/develop/src/NServiceBus.Transport.RabbitMQ/RabbitMQTransportInfrastructure.cs#L43

If either of those are true, then it's safe to go ahead and go into full-native mode. I had originally detected those and threw as well, but it seemed odd to know it was OK to turn off the TM but force the user to add a config setting.

Based on that, there's a case to be made that the `DisableTimeoutManager` API isn't needed at all, and you could instead just rely on the `DisableFeature` API. I added `DisableTimeoutManager` to RabbitMQ before realizing this. It's probably better to add it this point to have a consistent API across transports though...

bording:
Oh, and we currently have a PR in Rabbit to remove the `EnableByDefault` aspect of the TM behavior feature: https://github.com/Particular/NServiceBus.RabbitMQ/pull/393, but I'm still not convinced of the overall use case for needing to make the change.

bording:
One more thing, you should be able to set this to true: https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/develop/src/AcceptanceTests/TestSuiteConstraints.cs#L10

Scooletz:
Wow. Thank you @bording for such an insightful review! Going through point by point 👍

Scooletz:
Last commit ☝️ addresses this two:
- [x] Ordering of delivery constraints: `DoNotDeliverBefore`, `DelayDeliveryWith`
- [x] Throw on TTBR with a delayed delivery

Scooletz:
Last commits ☝️ addresses these:
- Delayed Retry handling
- Ordering of delivery constraints: `DoNotDeliverBefore`, `DelayDeliveryWith`
- Throw on TTBR with a delayed delivery

Scooletz:
Last commits ☝️ addresses these:
- Delayed Retry handling
- Ordering of delivery constraints: `DoNotDeliverBefore`, `DelayDeliveryWith`
- Throw on TTBR with a delayed delivery
- Throw on external timeout

Scooletz:
It looks to me that I addressed all the points
Ping @bording Could we switch to a regular review Brandon? It'd be easier for me to address the issues (if any occur :P )

src/Transport/DelayDelivery/NativeDelayDelivery.cs
```diff
@@ -26,39 +29,93 @@ public Task Init()
...
+            }
+
+            return StartupCheckResult.Success;
+        }
+
         static TimeSpan? GetVisbilityDelay(IOutgoingTransportOperation operation)
         {
-            var deliveryConstraint = operation.DeliveryConstraints.FirstOrDefault(d => d is DelayedDeliveryConstraint);
+            var constraints = operation.DeliveryConstraints;
+            var deliveryConstraint = constraints.FirstOrDefault(d => d is DelayedDeliveryConstraint);
```
bording: This logic still isn't correct. using only the first one found ignores the case of more than one being set, which isn't currently prevented by core.

src/Transport/DelayDelivery/NativeDelayDelivery.cs
```diff
@@ -26,39 +29,93 @@ public Task Init()
...
                 var delay = deliveryConstraint as DelayDeliveryWith;
                 if (delay != null)
                 {
                     value = delay.Delay;
                     return value;
                 }
-                var exact = deliveryConstraint as DoNotDeliverBefore;
-                if (exact != null)
+
+                if (constraints.Any() && constraints.Any(d => d is DiscardIfNotReceivedBefore))
```
bording: This doesn't appear to be correct either because if `DelayDeliveryWith` is set, the method would return before this is run.

src/Transport/DelayDelivery/NativeDelayDelivery.cs
```diff
@@ -26,39 +29,93 @@ public Task Init()
...
 
+        public static StartupCheckResult CheckForInvalidSettings(ReadOnlySettings settings)
+        {
+            var externalTimeoutManagerAddress = settings.GetOrDefault<string>("NServiceBus.ExternalTimeoutManagerAddress") != null;
+            
+            if (externalTimeoutManagerAddress)
+            {
+                return StartupCheckResult.Failed("An external timeout manager address cannot be configured because the timeout manager is not being used for delayed delivery.");
+            }
+
```
bording: This is missing the case where the TM isn't active but it's for a reason other than the user has specifically requested it.

This is required if you don't add the suggested checks to the `AzureStorageQueueInfrastructure` ctor, but even if you do, I recommend adding it anyway as a protection against running in a broken state.

src/Transport/AzureStorageQueueInfrastructure.cs
```diff
@@ -16,36 +16,31 @@
...
     {
-        internal AzureStorageQueueInfrastructure(ReadOnlySettings settings, string connectionString)
+        internal AzureStorageQueueInfrastructure(SettingsHolder settings, string connectionString)
         {
             this.settings = settings;
             this.connectionString = connectionString;
             serializer = BuildSerializer(settings);
 
-            timeoutsTableName = settings.GetOrDefault<string>(WellKnownConfigurationKeys.NativeTimeoutsTableName);
+            delayedDeliverySettings = settings.GetOrCreate<DelayedDeliverySettings>();
```
bording: I recommend checking to see if the TM is disabled or if the endpoint is send-only and setting `delayedDeliverySettings.TimeoutManagerDisabled` to true if they are. This prevents you ending up in a state where you know the TM is not going to be running, but then having to throw and tell the user to add a call to `DisableTimeoutManager`.

src/Transport/DelayDelivery/NativeDelayDelivery.cs
Scooletz: I reread the core's code for this. It's fixed in the recent one. I'm thinking: should I provide a test that breaks when one of these is reordered?

src/Transport/AzureStorageQueueInfrastructure.cs
Scooletz: Should this be a case? If they disabled TM they might have altered something else. Is setting the flag on our own ok? If I don't get it, pls rephrase.

src/Transport/DelayDelivery/NativeDelayDelivery.cs
Scooletz: If this is connected to the ctor, should this be the case?

src/Transport/DelayDelivery/NativeDelayDelivery.cs
Scooletz: Fixed in the recent PR

Scooletz:
Ping @brandon @danielmarbach 
The additional checks for various TM states were added. Please review again.

bording:
@Scooletz FYI I didn't get a chance to look over your changes today, but I will do so tomorrow!

src/AcceptanceTests/ConfigureAzureStorageQueueTransport.cs
```diff
@@ -47,7 +51,22 @@ public Task Configure(string endpointName, EndpointConfiguration configuration,
 
         return Task.FromResult(0);
     }
-    
+
+    static void EnableNativeDelayedDelivery(string endpointName, TransportExtensions<AzureStorageQueueTransport> transportConfig)
+    {
+        var delayedDelivery = transportConfig.DelayedDelivery();
+
+        delayedDelivery.DisableTimeoutManager();
```
bording: Since you're calling `DisableTimeoutManager` in your acceptance tests, you should also adjust your [TestSuiteConstraints](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/develop/src/AcceptanceTests/TestSuiteConstraints.cs#L10) to indicate that you do support native deferral.

src/Transport/PreventRoutingMessagesToTimeoutManager.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace NServiceBus.AzureStorageQueues
+{
+    using Features;
+
+    class PreventRoutingMessagesToTimeoutManager : Feature
+    {
+        public PreventRoutingMessagesToTimeoutManager()
+        {
+            EnableByDefault();
```
bording: FYI for RabbitMQ we currently have https://github.com/Particular/NServiceBus.RabbitMQ/issues/394 and https://github.com/Particular/NServiceBus.RabbitMQ/pull/393 discussing changing how this feature gets enabled to prevent it from being enabled in scenarios where it might not be desired.

src/Transport/AzureStorageQueueInfrastructure.cs
```diff
@@ -84,7 +90,7 @@ public override TransportReceiveInfrastructure ConfigureReceiveInfrastructure()
                     return new MessagePump(receiver, addressing, degreeOfReceiveParallelism);
                 },
                 () => new AzureMessageQueueCreator(client, GetAddressGenerator(settings)),
-                () => Task.FromResult(StartupCheckResult.Success)
+                () => Task.FromResult(NativeDelayDelivery.CheckForInvalidSettings(settings))
```
bording: The delay delivery check should be on `TransportSendInfrastructure` not `TransportReceiveInfrastructure` because you want the settings checked for send-only endpoints as well.

src/Transport/PreventRoutingMessagesToTimeoutManager.cs
Scooletz: I skimmed through these and it looks to me that this is a change that can be addressed on the core level.

Scooletz:
@danielmarbach Could you also review it?

src/Transport/DelayDelivery/NativeDelayDelivery.cs
```diff
@@ -25,40 +29,101 @@ public Task Init()
...
+            var delay = GetVisbilityDelay(constraints);
+            if (delay != null)
             {
-                return true;
+                if (FirstOrDefault<DiscardIfNotReceivedBefore>(constraints) != null)
+                {
+                    throw new Exception($"Postponed delivery of messages with TimeToBeReceived set is not supported. Remove the TimeToBeReceived attribute to postpone messages of type '{operation.Message.Headers[Headers.EnclosedMessageTypes]}'.");
+                }
+
+                await ScheduleAt(operation, UtcNow + delay.Value, cancellationToken).ConfigureAwait(false);
```
danielmarbach: If `ScheduleAt` would return `Task<bool>` we could turn this into a return, thoughts? Of course this would also influence timeouts.ExecuteAsync to return a bool

src/Transport/DelayDelivery/NativeDelayDelivery.cs
```diff
@@ -70,5 +135,26 @@ Task ScheduleAt(UnicastTransportOperation operation, DateTimeOffset date, Cancel
             timeout.SetOperation(operation);
             return timeouts.ExecuteAsync(TableOperation.Insert(timeout), cancellationToken);
         }
+
+        static TDeliveryConstraint FirstOrDefault<TDeliveryConstraint>(List<DeliveryConstraint> constraints)
```
danielmarbach: 👍

src/Transport/DelayDelivery/TimeoutsPoller.cs
```diff
@@ -213,7 +213,7 @@ UnicastTransportOperation CreateOperationForErrorQueue(UnicastTransportOperation
             errorQueue = settings.ErrorQueueAddress();
             var account = CloudStorageAccount.Parse(connectionString);
             table = await TimeoutEntity.BuiltTimeoutTableWithExplicitName(connectionString, tableName, cancellationToken).ConfigureAwait(false);
-            var container = account.CreateCloudBlobClient().GetContainerReference(table.Name.ToLower()); // TODO: can it be lowered?
+            var container = account.CreateCloudBlobClient().GetContainerReference(table.Name.ToLower());
```
danielmarbach: TimeoutEntity.BuiltTimeoutTableWithExplicitName accepts a `tableName` and then returns a table which we then use to get the container reference over `table.Name.ToLower()`. Should we consistently lower the table name in the settings and then always use the table name that is provided via settings?

src/Tests/NServiceBus.Azure.Transports.WindowsAzureStorageQueues.approved.cs
```diff
@@ -34,96 +35,16 @@ public static NServiceBus.TransportExtensions<NServiceBus.AzureStorageQueueTrans
...
         public static NServiceBus.TransportExtensions<NServiceBus.AzureStorageQueueTransport> UseSha1ForShortening(this NServiceBus.TransportExtensions<NServiceBus.AzureStorageQueueTransport> config) { }
     }
-    [System.ObsoleteAttribute("This class was replaced by extension methods on endpointConfiguration.UseTranspor" +
-        "t<AzureStorageQueue>(). Will be removed in version 8.0.0.", true)]
-    public class static ConfigureAzureMessageQueue { }
+    public class DelayedDeliverySettings
+    {
+        public DelayedDeliverySettings() { }
+        public void DisableTimeoutManager() { }
+        public void TableName(string timeoutTableName) { }
```
danielmarbach: So the tableName is optional correct?

src/Transport/AzureStorageQueueInfrastructure.cs
```diff
@@ -16,36 +17,41 @@
...
-            if (useNativeTimeouts)
+            var timeoutManagerFeatureDisabled = settings.GetOrDefault<FeatureState>(typeof(TimeoutManager).FullName) == FeatureState.Disabled;
+            var sendOnlyEndpoint = settings.GetOrDefault<bool>("Endpoint.SendOnly");
+
+            if (timeoutManagerFeatureDisabled || sendOnlyEndpoint)
             {
-                contraints.Add(typeof(DelayedDeliveryConstraint));
-                delayedDelivery = new NativeDelayDelivery(connectionString, timeoutsTableName);
-                shouldDispatch = delayedDelivery.ShouldDispatch;
+                // TM is automatically disabled to do not throuw during check
```
Scooletz: Will fix

src/Transport/AzureStorageQueueInfrastructure.cs
```diff
@@ -125,14 +131,14 @@ static MessageWrapperSerializer BuildSerializer(ReadOnlySettings settings)
...
-            return new TransportSendInfrastructure(BuildDispatcher, () => Task.FromResult(StartupCheckResult.Success));
+            return new TransportSendInfrastructure(BuildDispatcher, () => Task.FromResult(NativeDelayDelivery.CheckForInvalidSettings(settings)));
         }
 
         Dispatcher BuildDispatcher()
         {
             var addressing = GetAddressing(settings, connectionString);
             var addressRetriever = GetAddressGenerator(settings);
-            return new Dispatcher(addressRetriever, addressing, serializer, shouldDispatch);
+            return new Dispatcher(addressRetriever, addressing, serializer, delayedDelivery.Value.ShouldDispatch);
```
Scooletz: I think it's a leftover from an earlier stage. Let me investigate it.

src/Transport/DelayDelivery/NativeDelayDelivery.cs
```diff
@@ -25,40 +29,101 @@ public Task Init()
 
         public async Task<bool> ShouldDispatch(UnicastTransportOperation operation, CancellationToken cancellationToken)
         {
-            var delay = GetVisbilityDelay(operation);
-            if (delay == null)
+            var constraints = ((IOutgoingTransportOperation)operation).DeliveryConstraints;
```
Scooletz: Is is not. Will fix.

src/Transport/DelayDelivery/NativeDelayDelivery.cs
Scooletz: How could I transform it in a nice way to `Task<bool>`? It there an example of this?

src/Tests/NServiceBus.Azure.Transports.WindowsAzureStorageQueues.approved.cs
Scooletz: Nope. Configuration still uses this method to provide a required parameter. Should I change it? Not changing this leaves is more similar to other transports. Changing it - makes the requirement explicit moving it further from others.

src/Transport/DelayDelivery/TimeoutsPoller.cs
Scooletz: We could as:

> Table names are case-insensitive.

and

> All letters in a container name must be lowercase.

So, should we?

Scooletz:
Ping @danielmarbach

src/Tests/NServiceBus.Azure.Transports.WindowsAzureStorageQueues.approved.cs
danielmarbach: I think I already expressed my opinion. Happy to restate. I prefer to have it explicit in the API instead of throwing a runtime exception.

src/Transport/DelayDelivery/TimeoutsPoller.cs
danielmarbach: I'd say yes then you can lower it when the settings set it which I prefer.

src/Transport/DelayDelivery/NativeDelayDelivery.cs
danielmarbach: Scratch it. I downloaded the code and realized that ExecuteAsync is the SDK call. So there is nothing to gain here.

src/Transport/DelayDelivery/NativeDelayDelivery.cs
```diff
@@ -25,40 +29,101 @@ public Task Init()
...
+
+            return null;
+        }
 
+        static TimeSpan? ToNullIfNegative(TimeSpan value)
+        {
             return value <= TimeSpan.Zero ? (TimeSpan?)null : value;
         }
 
+        static bool TryProcessDelayedRetry(IOutgoingTransportOperation operation, out UnicastTransportOperation operationToSchedule, out DateTimeOffset scheduleDate)
```
danielmarbach: What is the value add of this being an interface instead of `UnicastTransportOperation` directly?

danielmarbach:
@Scooletz made a few comments. I'll remove my "request changes" since I don't want to hold this up unnecessarily.

Scooletz:
Daniel, I applied your changes. I'm leaving this PR open for sake of #224 which is targeted for this branch

Scooletz:
@Particular/azure-maintainers with Brandon and Daniel approving the initial PR, and being given #225 applied at the top, can I ask for an additional review as a whole, and approval for finalizing Native Delayed Delivery?

src/Transport/Config/DelayedDeliverySettings.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    using System;
+    using System.Text.RegularExpressions;
+
+    /// <summary>
+    /// Configures native delayed delivery.
+    /// </summary>
+    public class DelayedDeliverySettings
+    {
+        internal string Name { get; private set; }
```
danielmarbach: Minor inconsistency nitpick. One is a property, the other a field. Only saw it now. Sorry. But we can address this outside this PR. I'll merge

PR #224, opened 2017-06-19
Timeout headers are encoded with UTF8 with length preambule

Scooletz:
Connects to #220

This PR changes encoding of headers from JSON based to a custom UTF8 with length preamble.

A dummy performance test just to show that we are not that much slower than JSON. Bytes of course will be encoded when transmitted etc etc. This is just to remove JSON.NET dep and to have a small and efficient enough serializer.

|Serializer | Time to serialize 100000 dummy non-empty records |
|---|---|
|This custom binary serializer | 00:00:00.2466783 |
|JSON.NET | 00:00:00.5265303|

### DIFF
--- a/src/Tests/DelayDelivery/HeadersEncoderTests.cs
+++ b/src/Tests/DelayDelivery/HeadersEncoderTests.cs
@@ -3,8 +3,11 @@
     using System;
     using System.Collections;
     using System.Collections.Generic;
+    using System.Diagnostics;
+    using System.IO;
     using System.Linq;
     using WindowsAzureStorageQueues.DelayDelivery;
+    using Newtonsoft.Json;
     using NUnit.Framework;
     using NUnit.Framework.Interfaces;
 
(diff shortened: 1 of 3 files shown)


Scooletz:

yvesgoeleven:
@Scooletz what will the migration path look like for users that come from asp?

Scooletz:
This is only for the new and shiny native delayed delivery in ASQ. Native ASQ timeouts = this storage. Old timeouts stored in ASP = ASP storage. The common migration path for Native Delayed Delivery is to "shut down the TimeoutManager when you have no more old timeouts". Does it answer your question @yvesgoeleven ?

src/Tests/DelayDelivery/HeadersEncoderTests.cs
```diff
@@ -0,0 +1,90 @@
...
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using System.IO;
+    using System.Linq;
+    using WindowsAzureStorageQueues.DelayDelivery;
+    using Newtonsoft.Json;
+    using NUnit.Framework;
+    using NUnit.Framework.Interfaces;
+
+    public class HeadersEncoderTests
```
danielmarbach: Does this require a `[TestFixture]` attribute?

src/Tests/DelayDelivery/HeadersEncoderTests.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        [Test,Explicit("This is perf test")]
+        public void ComparePerformance()
+        {
+            const int n = 100000;
+
+            var headers = new Dictionary<string, string>
+            {
+                {Headers.ContentType, "type"},
```
danielmarbach: Is this the standard formatting we use without spaces between the curlies?

src/Transport/DelayDelivery/HeadersEncoder.cs
```diff
@@ -0,0 +1,97 @@
+﻿namespace NServiceBus.Azure.Transports.WindowsAzureStorageQueues.DelayDelivery
+{
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Runtime.Serialization;
+    using System.Text;
+
+    class HeadersEncoder
```
danielmarbach: Could be made static

danielmarbach: Before I dive deeper into this. What is the benefit of doing it this way over taking in another JSonSerializer which is shipped as source package?

Scooletz: I was trying to make the `SimpleJSON` configured and after spending some time come with this idea, which took me much less lines and is more efficient than storing all of it in JSON. I can revisit `SimpleJSON` though. Are you saying between the lines that "custom serialization is not that good". Maybe after all I should have used just SimpleJSON

src/Tests/DelayDelivery/HeadersEncoderTests.cs
Scooletz: Not required. Any class with a method annotated with `[Test]` is a fixture.

src/Transport/DelayDelivery/HeadersEncoder.cs
danielmarbach: I think there is always a tradeoff between efficiency and the ability to read the headers. I'm not saying custom serialization is not good. I'm asking if we are making the right tradeoffs actually. We are able to saturate ASQ IO limits. Cause ASQ requires us to serialize the whole message including the headers wouldn't we want a readable format for the headers?

Scooletz: Just a note. This is for `TimeoutEntity` stored in the table for Native Delayed Delivery. Not the headers that are attached to messages sent via Azure Storage Queues.

Scooletz:
Closed in favor of #225

PR #226, opened 2017-06-22
Making both settings fields

danielmarbach:


danielmarbach:
I did this as an in place edit on GH. Not sure if it will work

PR #227, opened 2017-06-22
Moving develop back to 7.0

Scooletz:
This change moves `develop` back to 7.0. This will allow to ship Native Delayed Delivery in minor, reducing all the overhead connected with bumping up the major. After reviewing all the changes needed, I'd say that it's cost-wise to stay on minor for this one. Additionally, the rest of transports is also shipping as minor.


Scooletz:

Issue #228, opened 2017-06-23
Investigate: Use delivery constraint instead of TTBR header


SeanFeldman:
Investigate if usage of TTBR can cause message loss.

SeanFeldman:
Connects to https://github.com/Particular/NServiceBus/issues/4807

SeanFeldman:
**Update**: it doesn't.

SeanFeldman:
The question is answered. TTBR with ASQ will not cause a message loss.

PR #229, opened 2017-06-23
Simplifying TTBR handling

SeanFeldman:
Connects to #228 

### DIFF
--- a/src/Transport/Dispatcher.cs
+++ b/src/Transport/Dispatcher.cs
@@ -85,17 +85,11 @@
                 throw new InvalidOperationException($"TimeToBeReceived is set to more than 7 days (maximum for Azure Storage queue) for message type '{messageType}'.");
             }
 
-            // TimeToBeReceived was not specified on message - go for maximum set by SDK
-            if (timeToBeReceived == TimeSpan.MaxValue)
-            {
-                timeToBeReceived = null;
-            }
-
-            var wrapper = BuildMessageWrapper(operation, timeToBeReceived, queue);
-            await Send(wrapper, sendQueue).ConfigureAwait(false);
+            var wrapper = BuildMessageWrapper(operation, queue);
+            await Send(wrapper, sendQueue, timeToBeReceived).ConfigureAwait(false);
         }
 
-        Task Send(MessageWrapper wrapper, CloudQueue sendQueue)
+        Task Send(MessageWrapper wrapper, CloudQueue sendQueue, TimeSpan? timeToBeReceived)
         {
             CloudQueueMessage rawMessage;
             using (var stream = new MemoryStream())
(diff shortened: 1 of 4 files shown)


src/Transport/Dispatcher.cs
```diff
@@ -132,7 +126,6 @@ MessageWrapper BuildMessageWrapper(IOutgoingTransportOperation operation, TimeSp
                 CorrelationId = headers.GetValueOrDefault(Headers.CorrelationId),
                 Recoverable = operation.GetDeliveryConstraint<NonDurableDelivery>() == null,
                 ReplyToAddress = headers.GetValueOrDefault(Headers.ReplyToAddress),
-                TimeToBeReceived = timeToBeReceived ?? TimeSpan.MaxValue,
```
yvesgoeleven: This is a wire level change, must be validated that it has no impact on older versions

src/Transport/Dispatcher.cs
```diff
@@ -85,25 +85,19 @@ public Task Dispatch(TransportOperations outgoingMessages, TransportTransaction
                 throw new InvalidOperationException($"TimeToBeReceived is set to more than 7 days (maximum for Azure Storage queue) for message type '{messageType}'.");
             }
 
-            // TimeToBeReceived was not specified on message - go for maximum set by SDK
-            if (timeToBeReceived == TimeSpan.MaxValue)
```
Scooletz: Was is a dead piece of code? What was it guarding against? As far as I see it, this calculation is not connected with TTBR header

src/Transport/MessageWrapper.cs
```diff
@@ -14,6 +14,7 @@ public class MessageWrapper : IMessage
 
         public string ReplyToAddress { get; set; }
 
+        [ObsoleteEx(Message = "Unnecessary property for MessageWrapper.", TreatAsErrorFromVersion = "8", RemoveInVersion = "9")]
```
Scooletz: By default, we use the core serializer to serialize message wrappers. When we remove this in v8, a following scenario could be considered.

- ASQ v9 sends a message to ASQ v8 endpoint
- ASQ v8 endpoint has its JSON serializer configured with `MissingMemberHandling.Error`. Because it contains no TTBR, deserialization throws

Is this something we should worry about?

SeanFeldman: Replace `ObsoleteEx` with just `Obsolete`+warning

SeanFeldman:
Changes as per discussion. Please review @yvesgoeleven @Scooletz

SeanFeldman:

src/Transport/Dispatcher.cs
SeanFeldman: The verification that `timeToBeReceived` is set to `null` when the original value is `TimeSpan.MaxValue` was already performed by the code executing a few lines earlier. So this check was unnecessary @Scooletz

Issue #230, opened 2017-06-26
ASQ requires always the timeout table name to be configured


Scooletz:
Connects to https://github.com/Particular/NServiceBus.AzureStorageQueues/issues/235

When trying to bump up docos for 7.4, which includes native delayed delivery, I smoke tested the ASQ sample. It failed with an exception pointing that the table for delayed delivery was not configured. This lead me to the [following line](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/develop/src/Transport/AzureStorageQueueInfrastructure.cs#L37) from infrastructure, which assumes the setting always being set:

```
  if (timeoutManagerFeatureDisabled || sendOnlyEndpoint)
            {
                // TM is automatically disabled to do not throw during check
                delayedDeliverySettings.DisableTimeoutManager();
            }
            
            delayedDelivery = new NativeDelayDelivery(connectionString, GetDelayedQueueTableName());
}
```

When TM is disabled, this might be done on purpose and we should not require to have this setting set.

SeanFeldman:
Replaced by #237

PR #231, opened 2017-06-26
Native Delayed Delivery should be enabled only if opt-in

Scooletz:
Connects to #230 

This PR introduces both:
- a failing test for a scenario with a disabled `TimeoutManager` and failing endpoint
- fix for proper management of native delayed delivery in ASQ

### DIFF
--- a/src/AcceptanceTests/Sending/When_TimeoutManager_disabled.cs
+++ b/src/AcceptanceTests/Sending/When_TimeoutManager_disabled.cs
@@ -1,0 +1,85 @@
+﻿namespace NServiceBus.AzureStorageQueues.AcceptanceTests.Sending
+{
+    using System;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using Features;
+    using NServiceBus.AcceptanceTests;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NUnit.Framework;
+    using AcceptanceTesting.Customization;
+    using Configuration.AdvanceExtensibility;
+
+    public class When_TimeoutManager_disabled : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_startup_properly_and_send_regular_messages()
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<Sender>(b => b.When((session, c) => session.Send(new MyMessage { Id = c.TestRunId })))
+                .WithEndpoint<Receiver>()
+                .Done(c => c.WasCalled)
+                .Run()
+                .ConfigureAwait(false);
+
+            Assert.True(context.WasCalled, "The message handler should be called");
+        }
+
+       
+        public class Context : ScenarioContext
+        {
+            public bool WasCalled { get; set; }
+        }
+
+        public class Sender : EndpointConfigurationBuilder
+        {
+            public Sender()
+            {
+                EndpointSetup<DefaultServer>(cfg =>
+                {
+                    cfg.UseTransport<AzureStorageQueueTransport>();
+                    cfg.DisableFeature<TimeoutManager>();
+                    cfg.ConfigureTransport().Routing()
+                        .RouteToEndpoint(typeof(MyMessage), typeof(Receiver));
+                });
+            }
+        }
+        
+        public class Receiver : EndpointConfigurationBuilder
+        {
+            public Receiver()
+            {
+                EndpointSetup<DefaultServer>(cfg =>
+                {
+                    cfg.UseTransport<AzureStorageQueueTransport>();
+                    cfg.DisableFeature<TimeoutManager>();
+                    
+                    // override delayed delivery settings, to make them unset
+                    cfg.GetSettings().Set<DelayedDeliverySettings>(new DelayedDeliverySettings());
+                });
+            }
+
+            public class MyMessageHandler : IHandleMessages<MyMessage>
+            {
+                public Context TestContext { get; set; }
+
+                public Task Handle(MyMessage message, IMessageHandlerContext context)
+                {
+                    if (TestContext.TestRunId != message.Id)
+                    {
+                        return Task.FromResult(0);
+                    }
+
+                    TestContext.WasCalled = true;
+
+                    return Task.FromResult(0);
+                }
+            }
+        }
+
+        public class MyMessage : IMessage
+        {
+            public Guid Id { get; set; }
+        }
+    }
+}
--- a/src/Transport/AzureStorageQueueInfrastructure.cs
+++ b/src/Transport/AzureStorageQueueInfrastructure.cs
@@ -33,8 +33,12 @@
                 // TM is automatically disabled to do not throw during check
                 delayedDeliverySettings.DisableTimeoutManager();
             }
-            
-            delayedDelivery = new NativeDelayDelivery(connectionString, GetDelayedQueueTableName());
+
+            string tableName;
+            if (IsNativeDelayedDeliveryCongifured(out tableName))
+            {
+                delayedDelivery = new NativeDelayDelivery(connectionString, tableName);
+            }
         }
 
         public override IEnumerable<Type> DeliveryConstraints
@@ -44,7 +48,7 @@
                 yield return typeof(DiscardIfNotReceivedBefore);
                 yield return typeof(NonDurableDelivery);
 
-                if (delayedDeliverySettings.TimeoutManagerDisabled)
+                if (delayedDelivery != null)
                 {
                     yield return typeof(DoNotDeliverBefore);
                     yield return typeof(DelayDeliveryWith);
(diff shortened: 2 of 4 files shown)


Scooletz:
Ping @Particular/azure-maintainers for a review

Scooletz:
FYI @weralabaj @bording 
In the recent commit I changed the condition for @weralabaj Constraints`. It's based on disabling the TM flag.

src/Transport/AzureStorageQueueInfrastructure.cs
```diff
@@ -184,14 +203,16 @@ public override Task Stop()
             return poller != null ? poller.Stop() : TaskEx.CompletedTask;
         }
 
-        string GetDelayedQueueTableName()
+        bool IsNativeDelayedDeliveryCongifured(out string tableName)
```
bording: typo: Congifured -> Configured

src/AcceptanceTests/Sending/When_TimeoutManager_disabled.cs
```diff
@@ -0,0 +1,85 @@
...
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using Features;
+    using NServiceBus.AcceptanceTests;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NUnit.Framework;
+    using AcceptanceTesting.Customization;
+    using Configuration.AdvanceExtensibility;
+
+    public class When_TimeoutManager_disabled : NServiceBusAcceptanceTest
```
bording: A better name for this test would be `When_TimeoutManager_feature_disabled`

src/Transport/AzureStorageQueueInfrastructure.cs
```diff
@@ -138,7 +142,17 @@ Dispatcher BuildDispatcher()
         {
             var addressing = GetAddressing(settings, connectionString);
             var addressRetriever = GetAddressGenerator(settings);
-            return new Dispatcher(addressRetriever, addressing, serializer, delayedDelivery.ShouldDispatch);
+            
+            Func<UnicastTransportOperation, CancellationToken, Task<bool>> shouldSend;
+            if (delayedDelivery != null)
```
bording: For consistency, should the also be calling `IsNativeDelayedDeliveryCongifured`?

bording:
@Scooletz I left some review comments about the specific changes in the PR, but I'm a bit confused overall.

This PR title says you're trying to make native delayed delivery opt-in only, so that you won't be using it be default? If that's the case, then I don't think that's been done here at all.

[This](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/develop/src/Transport/AzureStorageQueueInfrastructure.cs#L31) check only makes sense if you are intending to have native delays enabled by default, since it enables them even if the user hasn't opted-in.

The startup check [here](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/develop/src/Transport/DelayDelivery/NativeDelayDelivery.cs#L65) isn't going to be correct either if you're defaulting using the timeout manager.

src/Transport/AzureStorageQueueInfrastructure.cs
Scooletz: `delayedDelivery` is set only when `IsNativeDelayedDeliveryCongifured` so it's transitive, so it's ok, isn't it?

Scooletz:
I addressed the minors from your comments, but will answer the major one tomorrow @bording

src/Transport/AzureStorageQueueInfrastructure.cs
weralabaj: Given that, can that be every null https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/9d9b66e653f653a3c5ed53a56b9f62e7784b4925/src/Transport/AzureStorageQueueInfrastructure.cs#L26?

Scooletz:

src/Transport/AzureStorageQueueInfrastructure.cs
```diff
@@ -184,14 +205,16 @@ public override Task Stop()
...
 
-        string GetDelayedQueueTableName()
+        bool IsNativeDelayedDeliveryConfigured(out string tableName)
         {
-            if (string.IsNullOrEmpty(delayedDeliverySettings.Name))
+            if (string.IsNullOrEmpty(delayedDeliverySettings?.Name))
             {
-                throw new Exception("Native delayed delivery feature requires configuring a table suffix.");
+                tableName = null;
+                return false;
```
yvesgoeleven: Can we add a warning for the user here (or use a default tablename?)

Scooletz: I'd not add a default tablename as this would be shared with other endpoints. I can log a warning here though.

danielmarbach: Can this even happen? If we require delayed delivery to be enabled and the method accepts a table name? In the other cases when it is not opt-in we wouldn't like to see a WARN or would we?

src/Transport/AzureStorageQueueInfrastructure.cs
```diff
@@ -131,14 +137,24 @@ static MessageWrapperSerializer BuildSerializer(ReadOnlySettings settings)
...
-            return new Dispatcher(addressRetriever, addressing, serializer, delayedDelivery.ShouldDispatch);
+
+            Func<UnicastTransportOperation, CancellationToken, Task<bool>> shouldSend;
+            if (delayedDelivery != null)
+            {
+                shouldSend = delayedDelivery.ShouldDispatch;
+            }
+            else
+            {
+                shouldSend = (_, __) => Task.FromResult(true);
```
danielmarbach: This is only invoked once right? If not we should cache Task.FromResult

Scooletz: It is not. Will fix.

src/Transport/AzureStorageQueueInfrastructure.cs
yvesgoeleven: the config requires a settings object now, not required to fill the tablename. https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/231/files#diff-22d843709db087c2887fc569319806ffR58

danielmarbach: @yvesgoeleven yes https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/d72afec8f591729c679ab5b5749fbc9b4b472561/src/Transport/Config/AzureStorageTransportExtensions.cs#L119

yvesgoeleven: Should do a Guard.AgainstNull in that method and make the DelayedDeliverySettings internal, then it can't happen...

Scooletz: I'm not following the scenario that you're talking about @yvesgoeleven The check is done underneath in `.TableName`. What can't happen when we internalize the settings? Could you provide either a test or a step by step description? I can't see it.

src/Transport/DelayDelivery/NativeDelayDelivery.cs
```diff
@@ -56,27 +54,6 @@ public Task Init()
 
         public static DateTimeOffset UtcNow => DateTimeOffset.UtcNow;
 
-        public static StartupCheckResult CheckForInvalidSettings(ReadOnlySettings settings)
```
bording: You are still going to need this check, because there are invalid combinations to look for in all three operation modes.

src/Transport/AzureStorageQueueInfrastructure.cs
```diff
@@ -23,18 +24,23 @@ internal AzureStorageQueueInfrastructure(SettingsHolder settings, string connect
             this.connectionString = connectionString;
             serializer = BuildSerializer(settings);
 
-            delayedDeliverySettings = settings.GetOrCreate<DelayedDeliverySettings>();
+            delayedDeliverySettings = settings.GetOrDefault<DelayedDeliverySettings>();
 
             var timeoutManagerFeatureDisabled = settings.GetOrDefault<FeatureState>(typeof(TimeoutManager).FullName) == FeatureState.Disabled;
             var sendOnlyEndpoint = settings.GetOrDefault<bool>("Endpoint.SendOnly");
 
             if (timeoutManagerFeatureDisabled || sendOnlyEndpoint)
```
bording: Enabling native delays automatically doesn't really make sense when you're requiring users to opt-in to native delays.

Scooletz: This is done only if `delayedDeliverySettings` were found (there `?.` operator). So if you did not configured `DelayedDeliverySettings`, you don't get TM disabled. Does it make sense @bording ?

bording: While that might be true, as I mentioned in my overall review comment above, I don't like how subtle the logic is there. If you are in the default "only using the TM" mode, why are you even making that check in the first place?

Scooletz: I see. Let's ignore this. I'll provide a some kind of a global response to the major comment you provided.

SeanFeldman:
After reviewing this PR, discussing with other reviewers and TF, my conclusion is that the default behavior should be aligned with the one found in RabbitMQ rather than SQL. SQL transport is more of an exception due to how credentials are managed and inability to auto-generate timeouts table.

Closing this PR that will be followed up by another PR to reflect the necessary changed to make native delayed delivery function as required.

MarcinHoppe:
@SeanFeldman Can we delete the branch or do we want to keep it for time time for reference?

SeanFeldman:
Left it intentionally for now. Wouldn't worry to much about it. We've got enough folks with OCD that if it forgotten, they will remind me.

Guess who's one of them?! 😆

PR #244, opened 2017-08-03
[WIP] Hotfix 7.4.1

### LINKED ISSUE #241: BackoffStrategy is shared across threads which might result in less than optimal backoff behavior
### Who's affected

Anyone running Azure Storage Queues transport version 7.x running with a high [DegreeOfReceiveParallelism](https://docs.particular.net/transports/azure-storage-queues/configuration#configuration-parameters-degreeofreceiveparallelism).

### Symptoms

The back off strategy implemented by the transport did not respect parallel execution of message receive operations which could result in either backing off when it was not necessary, or not backing off when it was required. This would **not** cause endpoints to malfunction, but could result in a slightly higher number of operations against the Storage account or less than optimal storage querying. 

### PoA:

- [x] List all affected [supported versions](https://docs.particular.net/nservicebus/upgrades/supported-versions) ~~7.2, 7.3,~~ 7.4
- [x] Code changes --> [PR](https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/244)
- [ ] Doco update --> [PR](https://github.com/Particular/docs.particular.net/pull/3030)
  - [x] smoke test 
  - [x] load test 
- [ ] [Announcement](https://github.com/Particular/PlatformDevelopment/blob/master/guidelines/release-notification-template.md) --> [PR](https://github.com/Particular/Announcements/pull/173)
- [ ] [Release](https://github.com/Particular/PlatformDevelopment/blob/master/guidelines/branching-gitflow.md) -> link
- [ ] Announce on Google Group --> link
- [ ] Port to `develop` --> link

---
### Original issue

This is a bug but it does not affect correctness. It might result in unstable backoff behavior when running with high parallelism degree.

As received from @danielmarbach:
```
ASQ uses BackoffStrategy which uses internally timeToDelayNextPeek which is shared by multiple concurrent receivers.
```

backoff should be (probably) instantiated per receiving task and passed to every receiver call

SeanFeldman:
Fixes #241

Before the backup strategy was introduced the receiver used by the pump in parallel was designed to be stateless. By introducing the backup strategy in the constructor the receiver became stateful since the backup strategy has a field that needs to be updated. Before this change multiple parallel receive loops will try to update the same backup strategy and read/update the same field concurrently.

We decided against making the retriever per receive loop and just float in the backup strategy into the receiver to make it stateless again.

### DIFF
--- a/src/Transport/AzureMessageQueueReceiver.cs
+++ b/src/Transport/AzureMessageQueueReceiver.cs
@@ -8,12 +8,11 @@ namespace NServiceBus.AzureStorageQueues
 
     class AzureMessageQueueReceiver
     {
-        public AzureMessageQueueReceiver(IMessageEnvelopeUnwrapper unwrapper, CloudQueueClient client, QueueAddressGenerator addressGenerator, BackoffStrategy backoffStrategy)
+        public AzureMessageQueueReceiver(IMessageEnvelopeUnwrapper unwrapper, CloudQueueClient client, QueueAddressGenerator addressGenerator)
         {
             this.unwrapper = unwrapper;
             this.client = client;
             this.addressGenerator = addressGenerator;
-            this.backoffStrategy = backoffStrategy;
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


danielmarbach:
@Scooletz would you mind having a look at it, I seem to remember you and @tmasternak did the backup strategy stuff.

SeanFeldman:

Scooletz:
I don't mind @danielmarbach especially when it's been already merged :P 
A `BackoffStrategy` instance per parallel receiver would be the way I'd address it and it's the way how it has been addressed in this PR.

PR #245, opened 2017-08-04
BackoffStrategy per receiving tasks

danielmarbach:
Backport of #241 to develop


danielmarbach:
Will self-merge when green since already review by @boblangley

PR #247, opened 2017-08-16
Add .NET Standard 2.0 to ASQ + NServiceBus Core 7

SeanFeldman:
All projects compile.

Storage library has introduced some API changes between `net452` and `netstandard2.0`. Those are marked with `//TODO:` comments for especial review.

- [x] Review TODOs
- [x] follow the steps in https://github.com/Particular/DotNetCoreLaunch/issues/7


src/AcceptanceTests/Sending/When_delaying_messages_natively.cs
```diff
@@ -89,7 +89,7 @@ public class When_delaying_messages_natively : NServiceBusAcceptanceTest
             await delayedMessagesTable.ExecuteAsync(TableOperation.Insert(delayedMessageEntity)).ConfigureAwait(false);
         }
 
-        async Task<List<DynamicTableEntity>> GetDelayedMessageEntities()
+        async Task<IList<DynamicTableEntity>> GetDelayedMessageEntities()
```
danielmarbach: Not that it really matters but let's revert this change

src/Transport/DelayDelivery/NativeDelayDelivery.cs
```diff
@@ -22,7 +22,10 @@ public NativeDelayDelivery(string connectionString, string delayedMessagesTableN
             delayedMessagesTable = CloudStorageAccount.Parse(connectionString).CreateCloudTableClient().GetTableReference(delayedMessagesTableName);
             // In the constructor to ensure we do not force the calling code to remember to invoke any initialization method.
             // Also, CreateIfNotExistsAsync() returns BEFORE the table is actually ready to be used, causing 404.
-            delayedMessagesTable.CreateIfNotExists();
+
+            // TODO: original code was calling delayedMessagesTable.CreateIfNotExists(); as it was not affected by the bug the async version had.
+            // In case async version still returns before table is created, add a small delay.
+            delayedMessagesTable.CreateIfNotExistsAsync().GetAwaiter().GetResult();
```
danielmarbach: Another option would be to fire off the task and assign it to a field. When one of the public methods of this class is called await the outcome of the task.

src/NServiceBus.AzureStorageQueues.TransportTests/NServiceBus.AzureStorageQueues.TransportTests.csproj
```diff
@@ -1,26 +1,29 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net452</TargetFramework>
+    <TargetFrameworks>net452;netcoreapp2.0</TargetFrameworks>
```
danielmarbach: So the reasoning is that everything that is not prod lib that client will target against has no benefit of being targetted against netstandard?

src/AcceptanceTests/Sending/When_delaying_messages_natively.cs
SeanFeldman: This was intentional. netstd has changed the returned result to be IList. To avoid unnecessary conditional compilation, method signature is changes. It's a test method @danielmarbach

src/Transport/DelayDelivery/NativeDelayDelivery.cs
SeanFeldman: First we need to verify that the bug in the async version that was there before is resolved.
If that's the case, then yes, a task-based approach with await will be good.

src/NServiceBus.AzureStorageQueues.TransportTests/NServiceBus.AzureStorageQueues.TransportTests.csproj
SeanFeldman: And also to be able to run as far as I understood.

src/AcceptanceTests/Sending/When_delaying_messages_natively.cs
SeanFeldman: BTW, the sentence reads funny..
> Not that it really matters but let's revert this change

If it doesn't really matter, why the proposition? 😜

danielmarbach: From a perf standpoint, it does not matter since it is a test. Didn't know that the lib version changed this.

SeanFeldman: Sadly, they did.
Just like they removed some method overloads forcing you to seed `nulls` all over the place.

SeanFeldman:
@Particular/azure-maintainers any ideas why this is failing on TC but not locally?

src/Tests/NServiceBus.AzureStorageQueues.Tests.csproj
```diff
@@ -1,36 +1,39 @@
...
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0" />
+    <PackageReference Include="NUnit3TestAdapter" Version="3.8.0-alpha1" />
+    <PackageReference Include="RhinoMocks" Version="3.6.1" NoWarn="NU1701" />
     <PackageReference Include="Particular.CodeRules" Version="0.2.0" />
-    <PackageReference Include="ApiApprover" Version="6.1.0-beta2" />
-    <PackageReference Include="PublicApiGenerator" Version="6.1.0-beta2" />
     <ProjectReference Include="..\Transport\NServiceBus.AzureStorageQueues.csproj" />
   </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net452'">
```
SeanFeldman: @core-v7 folks (@timbussmann @andreasohlund) in regards to my [question](https://github.com/Particular/DotNetCoreLaunch/issues/7#issuecomment-322947378) - the answer @andreasohlund provided [here](https://github.com/Particular/DotNetCoreLaunch/issues/7#issuecomment-322983255) is what I have in place, but it [fails](https://builds.particular.net/viewLog.html?buildId=227292&buildTypeId=NServiceBus_Transports_NServiceBusAzureStorageQueues_1Build) on the build server. Any ideas?

SeanFeldman:
**Update**
It's compiling, **but** not ready for multiple targets and **doesn't** have all the [bells and whistles](https://github.com/Particular/DotNetCoreLaunch/issues/7) required for completeness. 
Also, TC is a bit [broken](https://particularsoftware.slack.com/archives/C03GKSAGX/p1503014813000004).

andreasohlund:
>It's compiling, but not ready for multiple targets

What's not working when multi targeting?

SeanFeldman:
@andreasohlund sorry, I didn't communicate the right info.
Compilation is working fine. It's the testing that was an issue. I have talked to @kbaley about future options and have a way forward for now. 
Small success celebration - ASQ is targeting full fw and .netstandard 🙂

SeanFeldman:
@andreasohlund is [`Directory.Build.props`](https://github.com/Particular/NServiceBus/blob/f5ddc3145d37bc0e147bc1880b4266d5bb931456/src/Directory.Build.props) required?

andreasohlund:
Strictly speaking no but it paves the way for us to update them more easily
and it avoid duplication so I'd recommend using it

SeanFeldman:
In such case, the general list in
https://github.com/Particular/DotNetCoreLaunch/issues/7 should have the
entry.

andreasohlund:
There is already an entry for adding the build props into /src ?

SeanFeldman:
Damn, I copied the list and it was updated since then 😃 
Ok, will remove it and just follow the source.

src/AcceptanceTests/NServiceBus.AzureStorageQueues.AcceptanceTests.csproj
```diff
@@ -1,25 +1,25 @@
...
-    <PackageReference Include="System.Spatial" Version="5.8.2" />
-    <ProjectReference Include="..\Transport\NServiceBus.AzureStorageQueues.csproj" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0" />
+    <PackageReference Include="NUnit3TestAdapter" Version="3.8.0-alpha1" />
+
+    <!-- Force tests to run with the latest Storage library -->
+    <PackageReference Include="WindowsAzure.Storage" Version="8.*" />
+
+    <!-- Remove package is updated to work with .NET Standard 2.0 -->
+    <PackageReference Include="NServiceBus.Newtonsoft.Json" Version="1.2.0-alpha0017" NoWarn="NU1701" />
```
SimonCropp: this should now be `2.0.0-*`

SimonCropp: and it is now net standard compat

SeanFeldman:
@Particular/azure-maintainers I've pushed more changes to complete the upgrade to NSB7.

Please review the changes.

src/AcceptanceTests/NServiceBus.AzureStorageQueues.AcceptanceTests.csproj
```diff
@@ -1,25 +1,24 @@
...
-    <PackageReference Include="Newtonsoft.Json" Version="10.0.2" />
-    <PackageReference Include="NServiceBus.AcceptanceTesting" Version="6.3.3" />
-    <PackageReference Include="NServiceBus.AcceptanceTests.Sources" Version="6.3.3" />
-    <PackageReference Include="NServiceBus" Version="6.3.3" />
-    <PackageReference Include="NUnit" Version="3.7.1" />
-    <PackageReference Include="System.Spatial" Version="5.8.2" />
     <ProjectReference Include="..\Transport\NServiceBus.AzureStorageQueues.csproj" />
+    <PackageReference Include="NServiceBus.AcceptanceTests.Sources" Version="7.0.0-*" />
+    <PackageReference Include="NUnit" Version="3.*" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0" />
```
SimonCropp: i have been going `15.*`

src/Directory.Build.props
```diff
@@ -0,0 +1,35 @@
+<Project>
+
+  <ItemGroup>
+    <PackageReference Include="Terrajobst.PlatformCompat.Analyzers" Version="0.0.95-alpha" PrivateAssets="All" />
```
SimonCropp: i dont think these add enough value

src/Directory.Build.targets
```diff
@@ -0,0 +1,23 @@
...
+    <Authors>NServiceBus Ltd</Authors>
+    <Company>NServiceBus Ltd</Company>
+    <PackageLicenseUrl>http://particular.net/LicenseAgreement</PackageLicenseUrl>
+    <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
+    <Copyright>Copyright 2010-$([System.DateTime]::UtcNow.ToString(yyyy)) NServiceBus. All rights reserved</Copyright>
+    <PackageTags>nservicebus servicebus msmq cqrs publish subscribe</PackageTags>
+    <PackageIconUrl>http://s3.amazonaws.com/nuget.images/NServiceBus_32.png</PackageIconUrl>
+    <PackageProjectUrl>https://docs.particular.net/nuget/$(PackageId)</PackageProjectUrl>
+    <PackageOutputPath>..\..\nugets</PackageOutputPath>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
```
SimonCropp: u only want this in the projects that produce packages

src/NServiceBus.AzureStorageQueues.TransportTests/NServiceBus.AzureStorageQueues.TransportTests.csproj
```diff
@@ -1,26 +1,29 @@
...
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>$(SolutionDir)NServiceBus.snk</AssemblyOriginatorKeyFile>
-    <DebugType>full</DebugType>
-    <DebugSymbols>True</DebugSymbols>
-    <WarningLevel>4</WarningLevel>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
SimonCropp: put this in the directory settings

src/Transport/NServiceBus.AzureStorageQueues.csproj
```diff
@@ -13,24 +13,25 @@
...
-    <PackageReference Include="WindowsAzure.Storage" Version="8.0.0" />
-    <PackageReference Include="NServiceBus" Version="6.0.0" />
-    <PackageReference Include="Obsolete.Fody" Version="4.2.2" />
-    <PackageReference Include="System.Spatial" Version="5.8.2" />
-    <PackageReference Include="Fody" Version="2.1.0" />
-    <PackageReference Include="Particular.CodeRules" Version="0.2.0" />
-    <PackageReference Include="NuGetPackager" Version="0.6.3" />
-    <PackageReference Include="GitVersionTask" Version="4.0.0-beta0012" />
+    <PackageReference Include="NServiceBus" Version="7.0.0-*" />
+    <PackageReference Include="NServiceBus.Newtonsoft.Json" Version="[2.0.0-*, 3.0.0)" />
```
SimonCropp: @Particular/nservicebus-maintainers @Particular/azure-maintainers is there any concerns with having a dependency between `NServiceBus.AzureStorageQueues` and `NServiceBus.Newtonsoft.Json`?

src/Transport/NServiceBus.AzureStorageQueues.csproj
```diff
@@ -13,24 +13,25 @@
...
     <PackageReference Include="Microsoft.Data.Services.Client" Version="5.8.2" />
-    <PackageReference Include="WindowsAzure.Storage" Version="8.0.0" />
-    <PackageReference Include="NServiceBus" Version="6.0.0" />
-    <PackageReference Include="Obsolete.Fody" Version="4.2.2" />
-    <PackageReference Include="System.Spatial" Version="5.8.2" />
-    <PackageReference Include="Fody" Version="2.1.0" />
-    <PackageReference Include="Particular.CodeRules" Version="0.2.0" />
-    <PackageReference Include="NuGetPackager" Version="0.6.3" />
-    <PackageReference Include="GitVersionTask" Version="4.0.0-beta0012" />
+    <PackageReference Include="NServiceBus" Version="7.0.0-*" />
```
SimonCropp: this should be a range

src/Transport/NServiceBus.AzureStorageQueues.csproj
```diff
@@ -13,24 +13,25 @@
...
+  <PropertyGroup>
+    <PackageId>NServiceBus.Azure.Transports.WindowsAzureStorageQueues</PackageId>
+    <Description>NServiceBus Transport for the Windows Azure Storage Queues</Description>
+  </PropertyGroup>
+
   <ItemGroup>
-    <None Include="..\..\packaging\nuget\nservicebus.azurestoragequeues.nuspec" Link="nservicebus.azurestoragequeues.nuspec" />
-    <PackageReference Include="Janitor.Fody" Version="1.4.1" />
+    <!-- lock WindowsAzure.Storage to version 8.0.0 -->
+    <PackageReference Include="WindowsAzure.Storage" Version="8.0.0" />
```
SimonCropp: why isnt this a range?

src/Transport/NServiceBus.AzureStorageQueues.csproj
```diff
@@ -13,24 +13,25 @@
...
     <PackageReference Include="Microsoft.Azure.KeyVault.Core" Version="2.0.4" />
```
SimonCropp: i think Microsoft.Azure.KeyVault.Core, Microsoft.Data.OData and System.Linq.Queryable are all redundant

src/Transport/NServiceBus.AzureStorageQueues.csproj
```diff
@@ -13,24 +13,25 @@
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <OutputPath>..\..\binaries\</OutputPath>
```
SimonCropp: this is redundant

src/Directory.Build.props
SeanFeldman: This is a part of the template promoted [here](https://github.com/Particular/DotNetCoreLaunch/issues/7)

src/Directory.Build.targets
SeanFeldman: Not following. We do produce a single package, the transport package.

andreasohlund: Yes but this file applies to all projects causing the test projects to output nugets as well

See https://github.com/Particular/NServiceBus.Gateway/blob/develop/src/NServiceBus.Gateway/NServiceBus.Gateway.csproj#L11

src/NServiceBus.AzureStorageQueues.TransportTests/NServiceBus.AzureStorageQueues.TransportTests.csproj
SeanFeldman: > put this in the directory settings

Example please?

src/Transport/NServiceBus.AzureStorageQueues.csproj
SeanFeldman: Should be, will fix.

src/Transport/NServiceBus.AzureStorageQueues.csproj
SeanFeldman: Those are [dependencies](https://www.nuget.org/packages/WindowsAzure.Storage/8.0.0) for `WindowsAzure.Storage`. Can I drop those and NuGet will identify those?

src/Directory.Build.targets
SeanFeldman: @andreasohlund so alternatively, I could move that instruction specifically into the product.csproj file. Would that work?

andreasohlund: That is the preferred way to do it

src/Directory.Build.props
SeanFeldman: It might be generated false alarms, but aside from that, does it hurt to have it?

SimonCropp: well since we should be actually running our acc tests on the target platforms, any problematic APIs should be caught as part of that. so for me it is redundant

src/NServiceBus.AzureStorageQueues.TransportTests/NServiceBus.AzureStorageQueues.TransportTests.csproj
SimonCropp: https://github.com/Particular/NServiceBus.Wcf/blob/develop/src/Directory.Build.props#L4

SeanFeldman: Will move to `Directory.Build.props`

src/AcceptanceTests/NServiceBus.AzureStorageQueues.AcceptanceTests.csproj
```diff
@@ -1,25 +1,24 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net452</TargetFramework>
+    <TargetFrameworks>net452;netcoreapp2.0</TargetFrameworks>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <DebugSymbols>True</DebugSymbols>
     <WarningLevel>4</WarningLevel>
```
SimonCropp: `<WarningLevel>4</WarningLevel>` is the default and can be removed

src/NServiceBus.AzureStorageQueues.TransportTests/NServiceBus.AzureStorageQueues.TransportTests.csproj
```diff
@@ -1,26 +1,29 @@
...
-    <PackageReference Include="NServiceBus.Testing.Fakes.Sources" Version="6.3.3" />
-    <PackageReference Include="NServiceBus.TransportTests.Sources" Version="6.3.3" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0" />
+    <PackageReference Include="NUnit3TestAdapter" Version="3.8.0-alpha1" />
+    <PackageReference Include="NServiceBus.Testing.Fakes.Sources" Version="7.0.0-*" />
+    <PackageReference Include="NServiceBus.TransportTests.Sources" Version="7.0.0-*" />
     <PackageReference Include="Microsoft.Azure.KeyVault.Core" Version="2.0.4" />
-    <PackageReference Include="Microsoft.Data.Edm" Version="5.8.2" />
-    <PackageReference Include="Microsoft.Data.OData" Version="5.8.2" />
     <PackageReference Include="Microsoft.Data.Services.Client" Version="5.8.2" />
```
SimonCropp: should be a range?

src/Tests/NServiceBus.AzureStorageQueues.Tests.csproj
```diff
@@ -1,36 +1,43 @@
...
-    <PackageReference Include="WindowsAzure.Storage" Version="8.1.4" />
-    <PackageReference Include="System.Spatial" Version="5.8.2" />
+    <None Remove="APIApprovals.ApproveAzureStorageQueueTransport.approved.txt" />
+    <None Remove="APIApprovals.ApproveAzureStorageQueueTransport.received.txt" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0" />
+    <PackageReference Include="NUnit" Version="3.*" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0" />
```
SimonCropp: duplicate ref

src/Transport/NServiceBus.AzureStorageQueues.csproj
SimonCropp: > Can I drop those and NuGet will identify those? 

yep. and same for System.Linq.Queryable

src/AcceptanceTests/NServiceBus.AzureStorageQueues.AcceptanceTests.csproj
```diff
@@ -1,25 +1,24 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net452</TargetFramework>
+    <TargetFrameworks>net452;netcoreapp2.0</TargetFrameworks>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <DebugSymbols>True</DebugSymbols>
```
SimonCropp: `<DebugSymbols>True</DebugSymbols>` why ?

src/Transport/NServiceBus.AzureStorageQueues.csproj
SeanFeldman: If I do so, get an error:
```error NU1701: Package 'System.Linq.Queryable 4.0.0' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v2.0'. This package may not be fully compatible with your project.```

src/Directory.Build.props
SimonCropp: and my reason for disliking them is the perf impact. https://developercommunity.visualstudio.com/content/problem/67259/roslyn-analyzers-are-being-slow-msbuild-using-lots.html  https://stackoverflow.com/questions/36461174/how-do-i-disable-all-roslyn-code-analyzers

it is a perf impact u pay on every keypress, and sometime even when idle. and the loss of RAM. if i were to chose a time to run this type of Analysis it would be in a unit test.

src/AcceptanceTests/NServiceBus.AzureStorageQueues.AcceptanceTests.csproj
SeanFeldman: Don't know. Ask @SimonCropp 😛 https://github.com/Particular/NServiceBus.AzureStorageQueues/blame/4ca310067c521619550de5622a3b8d4b9bf75f56/src/AcceptanceTests/NServiceBus.AzureStorageQueues.AcceptanceTests.csproj

SimonCropp: i suspect that was a copy paste issue?

src/Directory.Build.props
SeanFeldman: 👍 I'll disable it for now.

src/Transport/NServiceBus.AzureStorageQueues.csproj
SeanFeldman: @SimonCropp let's chat about this one?

src/Transport/NServiceBus.AzureStorageQueues.csproj
SeanFeldman: Can't be a range. We want to test against 7.0.0 and not the latest minor/hotfix

PR #254, opened 2017-08-31
Swapping crypto providers to .NET Standard equivalent

SeanFeldman:


SeanFeldman:
Please hold with merging.
It looks like `Terrajobst.PlatformCompat.Analyzers` might have false positives and all of those classes are supported by .NET Standard.

SeanFeldman:
Good to review/merge @Particular/azure-maintainers

Issue #251, opened 2017-08-29
Remove shortening from the transport


SeanFeldman:
~~With .NET Standard, cryptography algorithms SHA1 and MD5 are deprecated and are not available.~~
We should take the same approach as with ASB in version 7.0 where sanitization is delegated to the customers.

This is a breaking change.

### PoA
- [x] Remove shortening and update API
- [x] Doco ([PR](https://github.com/Particular/docs.particular.net/pull/3189))
  * Custom sanitization sample
  * Upgrade guide

SeanFeldman:
This is an API change and needs to be done prior to beta. FYI @Particular/dotnetcore-launch-squad

DavidBoike:
@SeanFeldman What about `SHA1.Create()` and `MD5.Create()`? Using that in Raven.

SeanFeldman:
Those are fine as they are invoking whatever is available on the platform code is running on @DavidBoike .

SeanFeldman:
The reason for this change in the general description was misleading. Updated.

SeanFeldman:
**Update**
PR is merged, doco + sample + upgrade guide is required.

SeanFeldman:
Beta on myget, closing.

PR #255, opened 2017-09-06
Remove default shortening

SeanFeldman:
Connects to #251


SeanFeldman:
@Particular/azure-maintainers please review.

src/Transport/Shortener.cs
```diff
@@ -1,51 +0,0 @@
-﻿namespace NServiceBus.AzureStorageQueues
-{
-    using System;
-    using System.Security.Cryptography;
-    using System.Text;
-
-    class Shortener
```
danielmarbach: Should these functions be documented in the upgrade guide? I think they should

SimonCropp: @danielmarbach yeah on a call with @SeanFeldman today he said much of this code would be in the upgrade guide

SeanFeldman: W/o documentation/sample it would be pretty much breaking every single ASQ customer.
Documented and sampled, rest assure.

Issue #236, opened 2017-07-06
Transport address translation should be consolidated and performed by ASQ TransportInfrastructure 


SeanFeldman:
Current implementation of ASQ Infrastructure code has [transport address translation](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/a8584226a2c764e929b33680168da600141f3890/src/Transport/AzureStorageQueueInfrastructure.cs#L154) (`ToTransportAddress()`) that is not compatible with ASQ native service [queue naming constraints](https://docs.microsoft.com/en-us/rest/api/storageservices/Naming-Queues-and-Metadata?redirectedfrom=MSDN).

To compensate for that, the transport is using [QueueAddressGenerator](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/a8584226a2c764e929b33680168da600141f3890/src/Transport/QueueAddressGenerator.cs#L9) to perform the necessary translations address translations in several places:
- [`AzureMessageQueueCreator`](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/a8584226a2c764e929b33680168da600141f3890/src/Transport/AzureMessageQueueCreator.cs#L30) 
- [`AzureMessageQueueReceiver`](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/a8584226a2c764e929b33680168da600141f3890/src/Transport/AzureMessageQueueReceiver.cs#L48)
- [`Dispatcher`](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/a8584226a2c764e929b33680168da600141f3890/src/Transport/Dispatcher.cs#L60)

All these components receive addresses that have gone through `AzureStorageQueueInfrastructure.ToTransportAddress()` already, invoked by Core.

SeanFeldman:
This code was introduced in ASQ v7. Need to review and verify if messages in flight are affected or not.

SeanFeldman:
@danielmarbach could you please update the issue with the findings?

danielmarbach:
The shared queue and the instance specific queue are translated by the core

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/Receiving.cs#L24
https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/Receiving.cs#L29

[...]

PR #257, opened 2017-09-11
Transport address

danielmarbach:
Solves https://github.com/Particular/NServiceBus.AzureStorageQueues/issues/236

Currently targeted at master. Needs to be re-targeted towards hotfix 

- `QueueAddressGenerator` is reused over send, receive and queue creation which introduces more potential for caching but a bit more contention on the concurrent dictionary
- `ToTransportAddress` passes the created queue name address generator.

### DIFF
--- a/src/Transport/AzureStorageQueueInfrastructure.cs
+++ b/src/Transport/AzureStorageQueueInfrastructure.cs
@@ -41,6 +41,7 @@
             }
 
             delayedDelivery = new NativeDelayDelivery(connectionString, delayedDeliverySettings.TableName);
+            addressGenerator = new QueueAddressGenerator(settings);
         }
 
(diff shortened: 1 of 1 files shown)


danielmarbach:
@Particular/azure-maintainers thoughts?

SeanFeldman:
👍 on consolidating instance creating and keeping as a member on the transport infrastructure level.

danielmarbach:
@Particular/azure-maintainers are you OK with the proposed fix and should I move forward with releasing this?

SeanFeldman:
@danielmarbach why does it target `master`?

danielmarbach:
Because I did not yet want to "burn" a hotfix branch. I will create the branches and then retarget

PR #259, opened 2017-09-11
required package for runnign tests on .netcore

SeanFeldman:
missing nuget for `dotnet test` execution required for .netcore


SeanFeldman:
The failing test will be addressed by a separate PR #258.
Merging.

PR #260, opened 2017-09-12
Show test results in Linux

SeanFeldman:
Changes needed to run and report test results in Linux


SeanFeldman:
Update: connection string passed via environment variable on linux is broken, needs to be escaped.
That will be a change on TC side. Pending the fix to confirm.

PR #263, opened 2017-09-14
Remove default serializer override

### LINKED ISSUE #253: Dont override the default serializer
* a transport should not have an opinion on this
* since the newtonsoft serializer is now external it meas the AzureStorageQueues nuget now has two extra dependencies. newtonsoft and NServiceBus.Newtonsoft.Json 
* when moving to the new major of AzureStorageQueues we will silently be moving them from the internal to the external json serializer. the problem is these to serializers are not 100% compatible. https://docs.particular.net/nservicebus/serialization/newtonsoft#compatibility-with-the-core-json-serializer
* makes the AzureStorageQueues nuget more prone to extra releases, since any major release of newtonsoft and NServiceBus.Newtonsoft.Json will require that the AzureStorageQueues nuget also be released
* the current "hidden + defaulted"  approach hides apis that people should be aware of and use eg https://docs.particular.net/nservicebus/serialization/newtonsoft#usage-custom-settings
* uses the non-public api of the magic key `"MainSerializer"`. this key is free to change in any patch release of the core
* prone to side effects on initialisation order. eg `AzureStorageQueueTransport` sets the main serializer and then the user sets the main serialzier on the standard endpointconfig api. I assume endpointconfig wins. but it only wins due to the initialisation order that the core uses for the transport infrastructure and the serialization definition
* uses the non-public api `SettingsHolder.Merge` via reflection. which is prone to breaking in any patch release of the core

@boblangley modified on 2017-09-15 to add POA

## POA
- [x] Transport no longer has an opinion on the serializer PR #263 
- [x] Remove dependencies on Newtonsoft/NServiceBus.Newtonsoft.Json PR #263
- [ ] Update guide for transport should address move to external serializer

boblangley:
Fixes #253


src/Directory.Build.targets
```diff
@@ -11,6 +11,7 @@
     <PackageProjectUrl>https://docs.particular.net/nuget/$(PackageId)</PackageProjectUrl>
     <PackageOutputPath>..\..\nugets</PackageOutputPath>
     <TargetsForTfmSpecificContentInPackage>$(TargetsForTfmSpecificContentInPackage);IncludePDBsInPackage</TargetsForTfmSpecificContentInPackage>
+    <LangVersion>Latest</LangVersion>
   </PropertyGroup>
```
SeanFeldman: Is this right? Didn't we say to lock it to version 7?

boblangley: I missed the memo. I will fix it.

src/Transport/Guard.cs
```diff
@@ -27,5 +30,12 @@ public static void AgainstNegativeAndZero(string argumentName, TimeSpan value)
             }
         }
 
+        public static void AgainstUnsetSerializerSetting(SettingsHolder settings)
```
danielmarbach: Interesting approach. I would have made this an extension method on top of SettingsHolder and put it into a static class like

```
static class SerializerExtensions {
   public static void AssertMainSerializerSet(this SettingsHolder settings) {}
```

or something the like

danielmarbach: I just realized that those Guard classes will eventually be copy pasted into other projects. Should they remain from from NSB specific things?

SeanFeldman:
@boblangley ATTs need to register serializer by default to be able to run.

src/Transport/Guard.cs
boblangley: At first I did it that way, as we had discussed. After looking at it the "Guards" seemed thematically to be a better fit, since those methods are also throwing and performing sanity checks.

src/Transport/Guard.cs
```diff
@@ -27,5 +30,12 @@ public static void AgainstNegativeAndZero(string argumentName, TimeSpan value)
...
+        {
+            if (!settings.TryGet<Tuple<SerializationDefinition, SettingsHolder>>(AzureStorageQueueTransport.SerializerSettingsKey, out var _))
+            {
+                throw new Exception("Due to message size restrictions setting a serializer is mandatory. Use `endpointConfiguration.UseSerialization<T>();` to select a serializer. If you are upgrading, install the `NServiceBus.Newtonsoft.Json` NuGet package and consult the upgrade guide for further information.");
```
SeanFeldman: >  Use `endpointConfiguration.UseSerialization<T>();`

We've used both single quote (`'`) and backtick (\`) in the past. Should we try to be consistent and use one? I don't mind which one BTW :smiley:

danielmarbach:
@boblangley it looks like you forgot to stage the changes we did in ConfigureAsq for the acceptance tests

src/Transport/Guard.cs
boblangley: @SeanFeldman or @yvesgoeleven Do you have a preference?

Should I add an extension method for `SettingsHolder` instead of another `Guard` method?

```csharp
        public override TransportInfrastructure Initialize(SettingsHolder settings, string connectionString)
        {
            Guard.AgainstNull(nameof(settings), settings);
            Guard.AgainstNullAndEmpty(nameof(connectionString), connectionString);

            settings.AssertMainSerializerSet();
```

vs

https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/754d165bf65de593c0507e05ef3c177be229c975/src/Transport/AzureStorageQueueTransport.cs#L29-L34

SeanFeldman: Keep it as a part of `Guard`.

danielmarbach:
@boblangley I added CSharp 7 support for resharper inspections

SeanFeldman:
Going to see if the recent changes on TC have fixed the build.

PR #264, opened 2017-09-15
Allow delayed delivery polling to be disabled

### LINKED ISSUE #250: Disabling delayed delivery polling to save Storage costs
Endpoints that do not utilize delayed delivery (example [here](https://github.com/Particular/NServiceBus.AzureStorageQueues/issues/248)) would not poll delayed delivery infrastructure to save Storage costs. 

Should an endpoint start leveraging delayed delivery, the transport should throw an exception to let know that disabling delayed delivery should not take place.

Configuration could look like the following:

```c#
var transport = endpointConfiguration.UseTransport<AzureStorageQueueTransport>();
var delayedDelivery = transport.DelayedDelivery();
delayedDelivery.DisableDelayedDelivery()
```

### PoA

- [x] Make the setting mandatory - **PR** #264
- [x] Add to the upgrade guide **PR** [#3233](https://github.com/Particular/docs.particular.net/pull/3233)
- [x] Update doco/samples **PR** [#3233](https://github.com/Particular/docs.particular.net/pull/3233)

SeanFeldman:
Connects to #250 
Fixes #250 

Prevent unnecessary Storage polling when an endpoint doesn't require features based on delayed delivery.

### DIFF
--- a/src/AcceptanceTests/Sending/When_delaying_messages_and_delayed_delivery_is_disabled.cs
+++ b/src/AcceptanceTests/Sending/When_delaying_messages_and_delayed_delivery_is_disabled.cs
@@ -24,17 +24,9 @@
                         options.DelayDeliveryWith(delay);
                         options.RouteToThisEndpoint();
 
-                        try
-                        {
-                            return session.Send(new MyMessage(), options);
-                        }
-                        catch (Exception exception)
-                        {
-                            ctx.SendException = exception;
-                            return Task.FromResult(0);
-                        }
+                        return session.Send(new MyMessage(), options);
                     }))
-                    .Done(ctx => ctx.WasCalled)
+                    .Done(ctx => true)
                     .Run();
 
                 Assert.IsFalse(context.WasCalled, "Endpoint's handler should never be invoked.");
--- a/src/AcceptanceTests/Sending/When_delaying_messages_and_delayed_delivery_is_disabled.cs
+++ b/src/AcceptanceTests/Sending/When_delaying_messages_and_delayed_delivery_is_disabled.cs
@@ -7,7 +7,7 @@
     using NServiceBus.AcceptanceTests.EndpointTemplates;
     using NUnit.Framework;
 
-    public class When_deferring_a_message_natively_and_delayed_delivery_is_disabled : NServiceBusAcceptanceTest
+    public class When_delaying_messages_and_delayed_delivery_is_disabled : NServiceBusAcceptanceTest
     {
         [Test]
         public async Task Should_throw()
--- a/src/Transport/DelayDelivery/NativeDelayDelivery.cs
+++ b/src/Transport/DelayDelivery/NativeDelayDelivery.cs
@@ -4,6 +4,7 @@
     using System.Collections.Generic;
     using System.Threading;
     using System.Threading.Tasks;
+    using AzureStorageQueues.Config;
     using DelayedDelivery;
     using DeliveryConstraints;
     using Features;
@@ -59,7 +60,7 @@
         {
             var externalTimeoutManagerAddress = settings.GetOrDefault<string>("NServiceBus.ExternalTimeoutManagerAddress") != null;
             var timeoutManagerFeatureActive = settings.GetOrDefault<FeatureState>(typeof(TimeoutManager).FullName) == FeatureState.Active;
-            var timeoutManagerDisabled = settings.Get<DelayedDeliverySettings>().TimeoutManagerDisabled;
+            var timeoutManagerDisabled = settings.Get<bool>(WellKnownConfigurationKeys.DelayedDelivery.DisableTimeoutManager);
 
             if (externalTimeoutManagerAddress)
             {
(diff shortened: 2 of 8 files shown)


SeanFeldman:
@Particular/azure-maintainers please review.
I've added an ATT to verify it works as expected. That should cover it. In case you think there's more to cover in tests, please let me know.

danielmarbach:
Should this be targeted against release?

src/AcceptanceTests/Sending/When_delaying_messages_and_delayed_delivery_is_disabled.cs
```diff
@@ -0,0 +1,82 @@
...
+                        {
+                            return session.Send(new MyMessage(), options);
+                        }
+                        catch (Exception exception)
+                        {
+                            ctx.SendException = exception;
+                            return Task.FromResult(0);
+                        }
+                    }))
+                    .Done(ctx => ctx.WasCalled)
```
danielmarbach: Shouldn't this done condition be

```
.Done(ctx => ctx.SendException != null)
```

src/AcceptanceTests/Sending/When_delaying_messages_and_delayed_delivery_is_disabled.cs
```diff
@@ -0,0 +1,82 @@
...
+                    .Run();
+
+                Assert.IsFalse(context.WasCalled, "Endpoint's handler should never be invoked.");
+            }
+            catch (Exception exception)
```
danielmarbach: When the done condition is changed could we then remove the try catch and just do:

```
Assert.IsFalse(context.WasCalled, "Endpoint's handler should never be invoked.");
Assert.True(exception.Message.Contains("Cannot delay delivery of messages when TimeoutManager is disabled or there is no infrastructure support for delayed messages"));
```

src/Transport/DelayDelivery/NativeDelayDelivery.cs
```diff
@@ -62,7 +63,7 @@ public static StartupCheckResult CheckForInvalidSettings(ReadOnlySettings settin
         {
             var externalTimeoutManagerAddress = settings.GetOrDefault<string>("NServiceBus.ExternalTimeoutManagerAddress") != null;
             var timeoutManagerFeatureActive = settings.GetOrDefault<FeatureState>(typeof(TimeoutManager).FullName) == FeatureState.Active;
-            var timeoutManagerDisabled = settings.Get<DelayedDeliverySettings>().TimeoutManagerDisabled;
+            var timeoutManagerDisabled = settings.Get<bool>(WellKnownConfigurationKeys.DelayedDelivery.DisableTimeoutManager);
```
danielmarbach: This change was done for alignment purposes? So instead of using the DelayedDeliverySettings instance directly you introduced the keys, correct?

SeanFeldman:
Indeed @danielmarbach. Went against the default branch 🙂

src/Transport/DelayDelivery/NativeDelayDelivery.cs
SeanFeldman: Correct. `DelayedDeliverySettings` was instantiated in 2 places and required and with `settings.Get<>()` or `settings.GetOrDefault<>()` it would require an awkward call to pass in the settings.

src/AcceptanceTests/Sending/When_delaying_messages_and_delayed_delivery_is_disabled.cs
SeanFeldman: Can't do this - try/catch is what gives me the exception 😃

SeanFeldman: Ignore my previous comment. Was looking at the wrong `exception` 🙂 
Problem with that approach is that test explodes with 
```
System.InvalidOperationException : Cannot delay delivery of messages when TimeoutManager is disabled or there is no infrastructure support for delayed messages.
   at NServiceBus.ThrowIfCannotDeferMessageBehavior.Invoke
```

SeanFeldman:
@danielmarbach your [suggestion](https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/264#discussion_r139078526) cannot be applied. Do you still think the merge should be blocked?

danielmarbach:
@SeanFeldman no I definitely don't want to hold the PR hostage

src/AcceptanceTests/Sending/When_delaying_messages_and_delayed_delivery_is_disabled.cs
SeanFeldman: Addressed in https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/264/commits/0b9a1aa60156b0ebd9c422a5183efc9fdd768cf1

SeanFeldman:
@Particular/azure-maintainers I've addressed the request change and Linux build is broken for the whole TC. Review and merge if good?

PR #267, opened 2017-09-19
setting not present unless disabled

yvesgoeleven:


SeanFeldman:
Hmmm, it sneaked in PR #264. Thanks for catching that.

PR #272, opened 2017-09-22
Environment variable for connection string shared between Windows and Linux

SeanFeldman:
This PR is to allow tests on Linux to run.
Will require local environment variable for connection string to be renamed from 
`AzureStorageQueueTransport.ConnectionString` to 
`AzureStorageQueueTransport_ConnectionString` (`.` replaced by `_`).

@Particular/azure-maintainers FYI


danielmarbach:
Will you send in something similar for ASB?

danielmarbach:
@kbaley this sounds also like something all engineers should be aware of?

danielmarbach:
@kbaley never mind, see https://github.com/Particular/NServiceBus/pull/4970#issuecomment-331792504

SeanFeldman:
🙂 I only got to this notification now @danielmarbach 
ASB could continue as-is because we'll support dots and underscores and ASB is not running on Linux.

PR #278, opened 2017-10-04
Re-enable ignored tests

danielmarbach:
Since the core beta is internally released that contains https://github.com/Particular/NServiceBus/pull/4956 we should re-enable those tests that were previously ignored


danielmarbach:
@Particular/azure-maintainers no idea why this PR fails with ApprovalTests

SeanFeldman:
API approver failed on TC and I'll need to investigate what was the reason. Won't block this PR.

Issue #279, opened 2017-10-06
ToTransportAddress should use dash instead of dot for satellites


danielmarbach:
also remove the test sanitizer

yvesgoeleven:
Note: verify the old guid generation logic, to see if this could not lead to different guids if set by the user

SeanFeldman:
Closed via https://github.com/Particular/NServiceBus.AzureStorageQueues/commit/896a81233a3c08f45ca02e6a7129fa36671ea45b

SeanFeldman:
@yvesgoeleven the note that you had - how applicable it is? In v7 of the transport it would never work.
Please re-open the issue _and_ add context if you think we're missing something

PR #280, opened 2017-10-06
ToTransportAddress uses dash instead of dot for qualifier

danielmarbach:
Addresses #279

### DIFF
--- a/src/Transport/AzureStorageQueueInfrastructure.cs
+++ b/src/Transport/AzureStorageQueueInfrastructure.cs
@@ -183,7 +183,7 @@
 
             if (logicalAddress.Qualifier != null)
             {
-                queue.Append("." + logicalAddress.Qualifier);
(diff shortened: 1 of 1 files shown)


danielmarbach:

danielmarbach:
This should be a fairly safe switch to do right? Because the '.' based queues would anyway not work

SeanFeldman:
@danielmarbach agreed. It would throw at runtime.

yvesgoeleven:
Can we run a test with the backward compat sanitization logic https://docs.particular.net/transports/azure-service-bus/sanitization?version=asb_8#automated-sanitization-backward-compatibility-with-versions-6-and-below before and after applying this change and see if the generated guids are still the same?

SeanFeldman:
@yvesgoeleven I assume your question is covered by #282 and doesn't need to be answered here 🙂

PR #286, opened 2017-10-31
Project cleanup

bording:
This cleans up a few things related to moving to the new project system and Particular.Packaging.

It also adds SourceLink support.


SeanFeldman:
Don't think this change has anything to do with the failing tests. Looks like Core threw a surprise party again, like in https://github.com/Particular/NServiceBus.AzureServiceBus/pull/666.

PR #288, opened 2017-11-02
Fix native delay

### LINKED ISSUE #287: Exception when creating SendOnly endpoint (8.0.0-beta0005)
When starting a SendOnly endpoint you will get the following exception:

```
System.InvalidOperationException: Receive transaction mode isn't available since this endpoint is configured to run in send-only mode.
   at NServiceBus.ConsistencyGuarantees.TransactionModeSettingsExtensions.GetRequiredTransactionModeForReceives(ReadOnlySettings settings)
   at NServiceBus.Azure.Transports.WindowsAzureStorageQueues.DelayDelivery.DelayedMessagesPoller.Init(ReadOnlySettings settings)
   at NServiceBus.Azure.Transports.WindowsAzureStorageQueues.DelayDelivery.DelayedMessagesPoller.Start(ReadOnlySettings settings, CancellationToken cancellationToken)
   at NServiceBus.AzureStorageQueues.AzureStorageQueueInfrastructure.Start()`
```

danielmarbach:
Closes #287 

With the latest core beta changes, it is no longer possible to acquire the transport receive transaction mode for send-only endpoints. This PR switches to only acquire the tx mode for normal endpoints. Send-only endpoints behave in `isAtMostOnce = false` mode.

### DIFF
--- a/src/AcceptanceTests/Sending/When_delaying_messages_natively_with_send_only.cs
+++ b/src/AcceptanceTests/Sending/When_delaying_messages_natively_with_send_only.cs
@@ -1,0 +1,173 @@
+namespace NServiceBus.Azure.Transports.WindowsAzureStorageQueues.AcceptanceTests.Sending
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using AcceptanceTesting.Customization;
+    using Microsoft.WindowsAzure.Storage;
+    using Microsoft.WindowsAzure.Storage.Table;
+    using NServiceBus.AcceptanceTests;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NUnit.Framework;
+    using Testing;
+
+    public class When_delaying_messages_natively_with_send_only : NServiceBusAcceptanceTest
+    {
+        [SetUp]
+        public new async Task SetUp()
+        {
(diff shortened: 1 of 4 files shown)


danielmarbach:
I had to move the acquiring of the tx mode to the start method because of https://github.com/Particular/NServiceBus/issues/5065

yvesgoeleven:
All tests are broken because: `Constructor on type 
 NServiceBus.Unicast.Messages.MessageMetadataRegistry' not found.`

danielmarbach:
@yvesgoeleven waat? How is this related to my changes. Another core update in between?

yvesgoeleven:
@danielmarbach I have no idea, just stating what I see on build server.

yvesgoeleven:
@danielmarbach looks like the code is doing some reflection here; 

https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/fix-native-delay/src/TransportTests/ConfigureAzureStorageQueueTransportInfrastructure.cs#L24

probably that constructor changed in latest core

danielmarbach:
@yvesgoeleven nice that we found that out almost at the same time 👯‍♂️ 

Fix is there

danielmarbach:
@yvesgoeleven green 🎆

SeanFeldman:
We should add this to the upgrade guide/doco as this is a change from version 7.

danielmarbach:
@SeanFeldman can you elaborate? Are you saying we need to document the fact that delayed delivery behaves slightly different when in send only mode because it no longer takes the set transaction into account? 

I have an idea how to fix this then we don't need to document anything ;)

danielmarbach:
@SeanFeldman pushed

SeanFeldman:
👍

Issue #276, opened 2017-10-03
Rewrite tests that leverage 'BuildAnotherConnectionString'


yvesgoeleven:
[This method](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/27026434ad535e6dd0a3b6d9450fb1184a493f36/src/AcceptanceTests/Utils.cs#L38) does return another connectionstring (format), but it still points to the same account. Some of the tests leveraging it assume it's also another account. So these tests aren't actually testing anything

- ~~Create a resource set of new accounts~~
- ~~Create another environment variable for the second account~~
- ~~Assign environment variable with one of the new accounts~~
- [x] Rewrite method to use that new env variable
- [x] Update tests accordingly

SeanFeldman:
The project already has 2 environment variables for connection strings, leveraging the same connection strings pool (TC shared variable). Going to use that.

SeanFeldman:
Tests were updated to use 2 connection strings.
Everything works as expected.

PR #290, opened 2017-11-16
Refactoring tests with second namespace

SeanFeldman:
Connects to #276 

These tests were using a spoofed 2nd account. Now they use two separate accounts.

@danielmarbach the tests I didn't like are `When_sending_messages_with_mapped_account_names` and `When_sending_message_with_custom_reply_to`. IMO they are unnecessarily complex and mix multiple concerns. We don't have to refactor them right now, but probably should address it at some point. For example, I could probably refactor `When_sending_messages_with_mapped_account_names` to use an "audit spy" endpoint to remove the complexity of native dequing and JSON parsing. If it's worth it, I'll go ahead and make the change.


SeanFeldman:
@Particular/azure-maintainers please review/merge

Issue #292, opened 2017-11-28
Delayed Delivery Feature Does Not Disable and Spams Blob/Tables in Storage Account


afyles:
## Who's affected

Anyone using Azure Storage Queues transport with native delayed deliver and wants to turn delayed delivery off.

## Symptoms

There's no way to disable delayed delivery to save the costs of Storage operations. As a result of that, the poller keeps connecting to Storage service (tables and blobs).

## Original issue

We are using .NET Web API to send messages to a backend endpoint.  No matter what we do we cannot get the Delayed Delivery feature to disable and stop making thousands of calls to the storage account in an attempt to acquire leases.  We see the traffic via Fiddler and as dependency calls in AppInsights.  Our Web API startup looks like the following:

```
EndpointConfiguration endpointConfiguration = new EndpointConfiguration("ProductWebApiEndpoint");
            endpointConfiguration.DisableFeature<MessageDrivenSubscriptions>();
            endpointConfiguration.SendOnly();
            TransportExtensions<AzureStorageQueueTransport> transport = endpointConfiguration.UseTransport<AzureStorageQueueTransport>();
            transport.DelayedDelivery().DisableTimeoutManager();
            transport.ConnectionStringName("ProductStorage");
            IEndpointInstance endpointInstance = Endpoint.Start(endpointConfiguration).GetAwaiter().GetResult();
            ContainerBuilder builder = new ContainerBuilder();
            builder.RegisterInstance<IMessageSession>(endpointInstance);
            config.DependencyResolver = new AutofacWebApiDependencyResolver(builder.Build());
```

We've tried the latest stable and the latest beta packages.  Here is the stable versions:
Autofac 4.6.2
Autofac.WebApi2 4.1.0
NServiceBus 6.4.3
NServiceBus.Azure.Transports.WindowsAzureStorageQueues 7.4.2
WindowsAzure.Storage 8.4.0

We would like to just shut the feature off, but we can't figure out how.  Sample Fiddler response:

HTTP/1.1 409 There is already a lease present.
Content-Length: 221
Content-Type: application/xml
Server: Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0
x-ms-request-id: 5e7ee2bd-001e-0020-6073-68cde1000000
x-ms-version: 2017-04-17
Date: Tue, 28 Nov 2017 18:06:43 GMT

[...]

PR #294, opened 2017-11-30
only enable delayed message poller if delayed delivery can be used

yvesgoeleven:
Fixes https://github.com/Particular/NServiceBus.AzureStorageQueues/issues/292

### DIFF
--- a/src/Transport/AzureStorageQueueInfrastructure.cs
+++ b/src/Transport/AzureStorageQueueInfrastructure.cs
@@ -44,6 +44,9 @@
             addressGenerator = new QueueAddressGenerator(settings);
         }
 
+        bool DelayedDeliveryCanBeUsed() => delayedDeliverySettings.TimeoutManagerDisabled;
+
+
         static string GetDelayedDeliveryTableName(string endpointName)
         {
             byte[] hashedName;
@@ -64,7 +67,7 @@
                 yield return typeof(DiscardIfNotReceivedBefore);
                 yield return typeof(NonDurableDelivery);
 
-                if (delayedDeliverySettings.TimeoutManagerDisabled)
+                if (DelayedDeliveryCanBeUsed())
                 {
                     yield return typeof(DoNotDeliverBefore);
                     yield return typeof(DelayDeliveryWith);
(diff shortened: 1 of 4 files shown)


src/Transport/AzureStorageQueueInfrastructure.cs
```diff
@@ -44,6 +44,9 @@ internal AzureStorageQueueInfrastructure(SettingsHolder settings, string connect
             addressGenerator = new QueueAddressGenerator(settings);
         }
 
+        bool DelayedDeliveryCanBeUsed() => delayedDeliverySettings.TimeoutManagerDisabled;
```
SeanFeldman: This is off. I'll elaborate below why.

src/Transport/AzureStorageQueueInfrastructure.cs
```diff
@@ -185,9 +188,12 @@ public override Task Start()
         {
             var maximumWaitTime = settings.Get<TimeSpan>(WellKnownConfigurationKeys.ReceiverMaximumWaitTimeWhenIdle);
             var peekInterval = settings.Get<TimeSpan>(WellKnownConfigurationKeys.ReceiverPeekInterval);
-            poller = new DelayedMessagesPoller(delayedDelivery.Table, connectionString, BuildDispatcher(), new BackoffStrategy(maximumWaitTime, peekInterval));
-            nativeDelayedMessagesCancellationSource = new CancellationTokenSource();
-            poller.Start(settings, nativeDelayedMessagesCancellationSource.Token);
+            if (DelayedDeliveryCanBeUsed())
```
SeanFeldman: This is a change that will cause an undesired outcome.
While there will be no message loss, there will be a situation where delayed messages will never get delivered. I'll provide a test  to demosntrate the scenario and an alternative PR.

SeanFeldman: Also, if should be surrounding `maximumWaitTime` and `peekInterval ` to be consistent with `develop` changes. Those are not required if poller is not enabled.

SeanFeldman:
This PR disable native delivery poller when TimeoutManager (TM) is turned off.
Problem with this apprach manifests itself in hybrid mode, when TM is **not** disabled and delayed deliver is in place. What happens is:
- All delayed messages are sent using native delayed delivery.
- Older timeouts in-flights are retrieved with TM.

Because of that combination, poller will be disabled. Meaning that any outgoing defered messages will be sent out using native delay delivery, but will not be received as poller will not be running. This means poller should only be disabled when TM is disabled **and** delayed delivery is off.

yvesgoeleven:
Closing in favor of https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/296

PR #298, opened 2017-12-06
Prepare for final beta release

bording:


SeanFeldman:
@bording given #299 looks like this will need to be re-addressed

PR #300, opened 2017-12-07
hotfix-7.5.1

### LINKED ISSUE #299: Messages delayed natively are not received in hybrid mode
## Who's affected

Anyone using Azure Storage Queues transport version 7.5.0 with native delay delivery and TimeoutManager enabled.

## Symptoms

Delayed messages sent from endpoints capable of native delayed delivery are not received.

## Original issue

When running in hybrid mode, delayed messages are sent using native mechanism and received using TimeoutManager _or_ native delayed delivery poller.

With bug fix for #293 another bug was introduced for hybrid mode - messages sent using native delayed delivery are not received as [poller is disabled](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/ed7fa3bb605f1d859cbe20b19b5fc659ffa2b14c/src/Transport/AzureStorageQueueInfrastructure.cs#L198). Reason why poller to be disabled is a wrong condition used for both
- Determining if transport can support [delayed delivery constrains](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/ed7fa3bb605f1d859cbe20b19b5fc659ffa2b14c/src/Transport/AzureStorageQueueInfrastructure.cs#L45)
- And to [enable/disable the poller](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/ed7fa3bb605f1d859cbe20b19b5fc659ffa2b14c/src/Transport/AzureStorageQueueInfrastructure.cs#L198).

The condition is evaluating two things
1. TimeoutManager is **disable** _and_
1. DelayedDelivery is **not disabled**
https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/ed7fa3bb605f1d859cbe20b19b5fc659ffa2b14c/src/Transport/AzureStorageQueueInfrastructure.cs#L67-L69

While this is good for determining the right delivery constraint the transport should return as supported, it is breaking the hybrid mode where messages should be received from TimeoutManager _and_ by poller. Same condition should not be used.

The poller enabling/disabling should solely evaluate if DelayedDelivery is **not disabled**.

## Affected versions 

- V7 (7.5.0) - required a patch and port to v8
- V8 
    - Blocking Azure Storage Persistence `release-8.0` PR https://github.com/Particular/NServiceBus.Persistence.AzureStorage
    - Rework of ASQ `release-8.0` branch

yvesgoeleven:
fixes #299

### DIFF
--- a/src/Transport/AzureStorageQueueInfrastructure.cs
+++ b/src/Transport/AzureStorageQueueInfrastructure.cs
@@ -68,6 +68,8 @@
             settings.GetOrDefault<bool>(WellKnownConfigurationKeys.DelayedDelivery.DisableTimeoutManager)
             && settings.GetOrDefault<bool>(WellKnownConfigurationKeys.DelayedDelivery.DisableDelayedDelivery) == false;
 
+        bool PollerCanBeUsed() => settings.GetOrDefault<bool>(WellKnownConfigurationKeys.DelayedDelivery.DisableDelayedDelivery) == false;
+
 
         public override IEnumerable<Type> DeliveryConstraints
         {
@@ -195,7 +197,7 @@
 
         public override Task Start()
         {
-            if (DelayedDeliveryCanBeUsed())
+            if (PollerCanBeUsed())
             {
                 var maximumWaitTime = settings.Get<TimeSpan>(WellKnownConfigurationKeys.ReceiverMaximumWaitTimeWhenIdle);
                 var peekInterval = settings.Get<TimeSpan>(WellKnownConfigurationKeys.ReceiverPeekInterval);


andreasohlund:
Is there any way we can test this with a ATT?

yvesgoeleven:
@andreasohlund not with just the ATT framework, this condition only applies to endpoints that have are migrating from the core timeoutmanager to the native delayed delivery feature, so we would have to reconstruct such a condition somehow

andreasohlund:
@Particular/rabbitmq-transport-maintainers did you manage to test this whenn you introduced the native timeouts in Rabbit?

adamralph:
The TF for native delays in RabbitMQ was @WojcikMike @bording and @SeanFeldman

yvesgoeleven:
@andreasohlund looks like the ASP ATT's did detect it https://github.com/Particular/NServiceBus.Persistence.AzureStorage/pull/183#issuecomment-349873772

andreasohlund:
>looks like the ASP ATT's did detect it Particular/NServiceBus.Persistence.AzureStorage#183 (comment) (at least that is my assumption)

I guess thats good but can't that be replicated with something like the InMemory persistence so we get a ATT failure in the ASQ build?

yvesgoeleven:
@andreasohlund sounds possible, but I would have to investigate as I was not involved in building this feature. Shall I spin off an additional issue for that, or block this one?

andreasohlund:
Spinn off sounds most resonable? (assuming that this is pretty urgent?)

yvesgoeleven:
It is yes

The tests on the buildserver seem to be broken, haven't even started, immediate 'Process exited with code 1', could this have to do with the TC upgrade?

SeanFeldman:
.NET Core builds should not be executed, ignoring those and merging.

Issue #124, opened 2016-06-06
ASQ doesn't work with XML MessageWrappers


timbussmann:
When configuring the acceptance tests to work with the XmlSerializer (`.UseSerializer<XmlSerializer>()`) the `MessageWrapper` class won't be properly deserialized which makes the endpoint unable to receive messages.

It seems like the internal `XmlSerializerCache` isn't initialized with the `MessageWrapper` class. On endpoints which send messages this seems to work after they send a message since the serialization logic does add missing types to the cache.

@Particular/azure-storage-queues-maintainers ping. This seems to be must-have before the RC.

timbussmann:
note: I think we maybe need to handle this in the XmlSerializer. The serialization logic does initialize the type at runtime if it's missing in the cache while deserialization does not. We could adjust that behavior to also initialize types at runtime if they seem missing. /cc @Particular/nservicebus-maintainers

Scooletz:
It's highly unlikely that anyone is using XML instead of JSON, but as we provide a way to configure it, we should be handling it right. I'm looking forward to hearing from @Particular/nservicebus-maintainers if this is going to be handled in the core and ASQ will just bump up the version of the core.

timbussmann:
@Particular/nservicebus-maintainers @Particular/azure-storage-queues-maintainers I think I found a minimal invasive solution by sneaking in this config somewhere on the ASQ setup:
`settings.Get<Conventions>().AddSystemMessagesConventions(t => t == typeof(MessageWrapper));`

this is the simplest way of getting `MessageWrapper` being accepted as a message type which will also initialize it properly (it was skipped because it's from a particular assembly and we would exclude messages from particular assemblies). Will test whether that works and raise a PR if it does.

Scooletz:
Would be more than happy to review & accept it :)

timbussmann:
should be fixed with #125

PR #304, opened 2018-01-10
Remove MessageWrapper from Conventions and MessageMetadataRegistry

danielmarbach:
Related to https://github.com/Particular/ServiceControl.TransportAdapter/issues/23

- Registering the MessageWrapper in the conventions as a system message is not necessary because MessageWrapper implements IMessage
- Registering the MessageWrapper in the message metadata registry is not necessary because the XmlSerializer caches types during runtime when necessary

@Scooletz  and I did some digging and it looks like those changes were introduced for the acceptance tests around the 7.0.0 RC timeframe. When the core v6.0.0 final was released the XmlSerializer was already implemented with runtime caching of types where necessary. 

Related issue https://github.com/Particular/NServiceBus.AzureStorageQueues/issues/124
The related PR for the issue above https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/125
as well as https://github.com/Particular/NServiceBus.AzureStorageQueues/commit/e5968b8ebdafcfb8f610fa973d027ec6a2513757

We smoked tested these changes also against core v6.0.0 with messages already in the queue sent by an unpatched version of the ASQ transport. We did it for the 6.0.0 to cover the whole nuspec range. Everything works smoothly and the xml envelop when used with the XmlSerializer is still the same and thus backward compliant.

In addition to that we sent in a PR that showes that all acceptance tests still pass even with the XmlSerializer as a wrapper serializer

https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/305

### Before

`<?xml version="1.0"?><MessageWrapper xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns="http://tempuri.net/NServiceBus.Azure.Transports.WindowsAzureStorageQueues">...`

### After

`<?xml version="1.0"?><MessageWrapper xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns="http://tempuri.net/NServiceBus.Azure.Transports.WindowsAzureStorageQueues">`

### DIFF
--- a/src/Transport/AzureStorageQueueTransport.cs
+++ b/src/Transport/AzureStorageQueueTransport.cs
@@ -2,14 +2,12 @@ namespace NServiceBus
 {
     using System;
     using System.Reflection;
-    using Azure.Transports.WindowsAzureStorageQueues;
     using AzureStorageQueues;
     using MessageInterfaces;
     using Routing;
     using Serialization;
     using Settings;
     using Transport;
-    using Unicast.Messages;
 
     /// <summary>
     /// Transport definition for AzureStorageQueue
@@ -27,12 +25,6 @@ namespace NServiceBus
             Guard.AgainstNullAndEmpty(nameof(connectionString), connectionString);
             // configure JSON instead of XML as the default serializer:
             SetMainSerializer(settings, new JsonSerializer());
-
-            // register the MessageWrapper as a system message to have it registered in mappings and serializers
-            settings.GetOrCreate<Conventions>().AddSystemMessagesConventions(t => t == typeof(MessageWrapper));
-
-            // register metadata of the wrapper for the sake of XML serializer
-            settings.Get<MessageMetadataRegistry>().GetMessageMetadata(typeof(MessageWrapper));
 
             DefaultConfigurationValues.Apply(settings);
 


danielmarbach:
@timbussmann added you as a reviewer because our crime scene investigation showed traces of your involvement.

SzymonPobiega:
Great news! Now if we only added that API to ignore the serialization check the Transport Adapter would be very happy :-)

src/Transport/AzureStorageQueueTransport.cs
```diff
@@ -28,12 +26,6 @@ public override TransportInfrastructure Initialize(SettingsHolder settings, stri
             // configure JSON instead of XML as the default serializer:
             SetMainSerializer(settings, new JsonSerializer());
 
-            // register the MessageWrapper as a system message to have it registered in mappings and serializers
-            settings.GetOrCreate<Conventions>().AddSystemMessagesConventions(t => t == typeof(MessageWrapper));
```
yvesgoeleven: Test this with custom conventions, if IMessage is not considered will it still be picked up

danielmarbach: Will rerun the tests, thanks didn't think about that. Good catch

SeanFeldman:
> Great news! Now if we only added that API to ignore the serialization check the Transport Adapter would be very happy :-)

@SzymonPobiega setializer part is likely to stay (for ASQ and ASB both). At least the [workaround/hack](https://github.com/Particular/ServiceControl.TransportAdapter/issues/23#issuecomment-356398605) is not going to be that bad. Only setting the default serializer usef for `MessageWrapper` for ASQ and a fake one for ASB.

SeanFeldman:
Any idea why build is failing?

src/Transport/AzureStorageQueueTransport.cs
danielmarbach: Did the following:

Used the `AzureStorageQueues` sample, switched it to `XmlSerializer`, added unobtrusive mode and removed `IMessage` marker from messages in shared assembly
Started both endpoints 1 and 2, sent messages from endpoint 1
Stopped endpoint 2
Sent in more messages from endpoint 1
Stopped both endpoints
Deployed patched ASQ version to both endpoints
Started Endpoint 1
Sent more messages from endpoint 1
Started Endpoint 2 and all messages were received

All under the eagle eyes of Mr. @SeanFeldman

yvesgoeleven: 👍

timbussmann: @danielmarbach did you make sure your tests is really exposed to the original issue https://github.com/Particular/NServiceBus.AzureStorageQueues/issues/124. Also see the comment in that issue:

>  On endpoints which send messages this seems to work after they send a message since the serialization logic does add missing types to the cache.

so you would have to test this on endpoints which only receive messages.

but it also seems like there is quite a high chance that this has been fixed in the xml serializer as proposed in this comment:
> I think we maybe need to handle this in the XmlSerializer. The serialization logic does initialize the type at runtime if it's missing in the cache while deserialization does not. We could adjust that behavior to also initialize types at runtime if they seem missing

I'm not 100% sure though, so testing the above mentioned scenario might be a good idea.

danielmarbach: The sample has to endpoints. Endpoint 1 sends a message while endpoint 2 only receives and replies. It seems that scenario is covered right?

timbussmann: yes

PR #306, opened 2018-01-11
Hotfix 7.5.2

danielmarbach:
Remove unnecessary settings to simplify transport adapter scenarios.

- Registering the MessageWrapper in the conventions as a system message is not necessary because MessageWrapper implements IMessage
- Registering the MessageWrapper in the message metadata registry is not necessary because the XmlSerializer caches types during runtime when necessary

For more information see https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/304

### DIFF
--- a/src/Transport/AzureStorageQueueTransport.cs
+++ b/src/Transport/AzureStorageQueueTransport.cs
@@ -2,14 +2,12 @@ namespace NServiceBus
 {
     using System;
     using System.Reflection;
-    using Azure.Transports.WindowsAzureStorageQueues;
     using AzureStorageQueues;
     using MessageInterfaces;
     using Routing;
     using Serialization;
     using Settings;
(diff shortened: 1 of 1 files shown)


SeanFeldman:
@danielmarbach either a description section needs to be updated or a separate issue opened for release notes. Also, let's add an explicit step for smoke testing to run it through again once the bits are on MyGet. Sounds ok?

danielmarbach:
Is green of build servers but seems the github hook was effected due to the github outtake. Merging manually

SeanFeldman:
Hmm, build server still has it in queue...
image

danielmarbach:
Ups then I didn't look closely enough. I'm guessing it is not a big deal since the PR against the hotfix branch was green already. But time will tell

Issue #308, opened 2018-01-12
ASQ v7: API for secure connection string with default connection string only is broken


SeanFeldman:
## Who's affected

- Anyone using NServiceBus version 7 and above

## Symptoms

When using ASQ with [secured connection string](https://docs.particular.net/transports/azure-storage-queues/multi-storageaccount-support#cross-namespace-routing-using-aliases-instead-of-connection-strings) (alias instead of connection string) with a default storage account, dispatches are failing with `NullReferenceException`.

## Details

Default storage account connection string is not associated with the default alias. Instead, the alias is mapped to an empty string, which causes `CreateQueueClients` to build a null `CloudQueueClient` that is used during dispatching, causing an exception.

According to documentation, API should be used in the following manner:

```c#
var transport = endpointConfiguration.UseTransport<AzureStorageQueueTransport>();
transport.ConnectionString("account_A_connection_string");
transport.UseAccountAliasesInsteadOfConnectionStrings();
transport.DefaultAccountAlias("account_A");
```
In reality this is failing when an attempt is made to dispatch a message to `account_A`.

## Workaround

Workaround is to register default connection string and its alias via `AccountRouting()` API.

```c#
var transport = endpointConfiguration.UseTransport<AzureStorageQueueTransport>();
transport.ConnectionString("account_A_connection_string");
transport.UseAccountAliasesInsteadOfConnectionStrings();
transport.DefaultAccountAlias("account_A");
//workaround - force alias registration
transport.AccountRouting().AddAccount("account_A", "account_A_connection_string"); 
```

## Affected versions

- 7.4.x
- 7.5.x
- 8.0

SeanFeldman:
Documentation update is needed and then this issue could be closed.

SeanFeldman:
Closing in favour of documentation PR https://github.com/Particular/docs.particular.net/pull/3595

PR #310, opened 2018-01-17
[WIP] Issue 309 default alias

SeanFeldman:
Connects to #309 

**DO NOT MERGE**

This PR is intended to only demonstrate the issue.

Issue #309, opened 2018-01-16
UseAccountAliasesInsteadOfConnectionStrings with default connection string exposes raw connections strings


SeanFeldman:
## Who's affected

- Anyone using Azure Storage Queues transport version 7.5.x and above

## Symptoms

When using ASQ with default connection string and alias to secure it, messages sent with raw connection string instead of alias.

### Details

Azure Storage Queues transport (ASQ) has a feature to [secure connections strings](https://docs.particular.net/transports/azure-storage-queues/multi-storageaccount-support#cross-namespace-routing-using-aliases-instead-of-connection-strings).

> In order to prevent accidentally leaking connection string values, it is recommended to use aliases instead of raw connection strings. When applied, raw connection string values are replaced with registered aliases removing the possibility of leaking a connection string value. **When using a single account, aliasing connection string is limited to just enabling it.** When multiple accounts are used, an alias has to be registered for each storage account.

This appear to be broken when a single (default) connection string is used. A separate [bug issue is raised ](https://github.com/Particular/NServiceBus.AzureStorageQueues/issues/308).
Unfortunately, the API is broken in a way that leads customers to assume that the default connection string is secured by invoking the API, when in reality it is not.

When API is called

```c#
transport.UseAccountAliasesInsteadOfConnectionStrings();
transport.DefaultAccountAlias("account_A");
```

Sending messages will contain the raw connection string rather than its alias as per documentation.

```json
{
  "IdForCorrelation": null,
  "Id": "7427f558-9e38-4959-9561-a869017ea425",
  "MessageIntent": 1,
  "ReplyToAddress": "samples-azure-storagequeues-endpoint1@UseDevelopmentStorage=true",
  "Headers": {
    "NServiceBus.MessageId": "7427f558-9e38-4959-9561-a869017ea425",
    "NServiceBus.MessageIntent": "Send",
    "NServiceBus.ReplyToAddress": "samples-azure-storagequeues-endpoint1@UseDevelopmentStorage=true",
  },
  "Body": "...",
  "CorrelationId": "7427f558-9e38-4959-9561-a869017ea425",
  "Recoverable": true
}
```

### Affected versions

- ASQ 7.4.x
- ASQ 7.5.x
- ASQ 8

### Expected behaviour

ASQ should [behave as ASB](https://docs.particular.net/transports/azure-service-bus/securing-connection-strings#backward-compatibility) and provide a default alias, which is used for the default connection string when connection strings should be secured.

[...]

PR #311, opened 2018-01-17
Hotfix-7.5.3

SeanFeldman:
Connects to #309
Replaces PR #310 

Fixes bug #308 - When using ASQ with default connection string and alias to secure it, messages sent with raw connection string instead of alias.

### DIFF
--- a/src/AcceptanceTests/Configuration/When_using_alias_instead_of_connection_string_for_default_account.cs
+++ b/src/AcceptanceTests/Configuration/When_using_alias_instead_of_connection_string_for_default_account.cs
@@ -3,6 +3,7 @@
     using System.IO;
     using System.Threading.Tasks;
     using AcceptanceTesting;
+    using Azure.Transports.WindowsAzureStorageQueues;
     using Azure.Transports.WindowsAzureStorageQueues.AcceptanceTests;
     using EndpointTemplates;
     using Microsoft.WindowsAzure.Storage;
@@ -14,6 +15,7 @@
     public class When_using_alias_instead_of_connection_string_for_default_account : NServiceBusAcceptanceTest
     {
         CloudQueue destinationQueue;
+        const string EndpointName = nameof(SenderEndpoint);
 
         public When_using_alias_instead_of_connection_string_for_default_account()
         {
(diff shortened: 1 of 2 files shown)


SeanFeldman:

src/AcceptanceTests/Configuration/When_using_alias_instead_of_connection_string_for_default_account.cs
```diff
@@ -0,0 +1,77 @@
...
+    using Microsoft.WindowsAzure.Storage;
+    using Microsoft.WindowsAzure.Storage.Queue;
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Linq;
+    using NUnit.Framework;
+
+    public class When_using_alias_instead_of_connection_string_for_default_account : NServiceBusAcceptanceTest
+    {
+        CloudQueue destinationQueue;
+        const string EndpointName = nameof(SenderEndpoint);
```
danielmarbach: Shouldn't this be using the naming provided by the acceptance tests?

SeanFeldman: @danielmarbach I thought testing `Conventions` would turn `When_using_alias_instead_of_connection_string_for_default_account.SenderEndpoint` to a queue name and that would be too long, requiring transport to sanitize the name and test to know what that value would be, but as it turns out, convention takes just the endpoint class name w/o the parent class. Awesome!

SeanFeldman: Turns out that's the case when running locally, but not on TC 🙁 It [fails on TC](https://builds.particular.net/viewLog.html?buildId=269658&tab=buildResultsDiv&buildTypeId=NServiceBus_Transports_NServiceBusAzureStorageQueues_31TestNetFrameworkOnWindows) because queue name is too long and is getting sanitized. I'll look into it.

SeanFeldman: My previous comment was inaccurate. It happened because I was trying to get the queue name too early in the ATT lifecycle. Problem is what I've originally had - endpoint is too long. The transport is going to sanitize it. I'll raise a separate issue for sanitization.

Issue #312, opened 2018-01-17
UseAccountAliasesInsteadOfConnectionStrings with default connection string exposes raw connections strings


SeanFeldman:
## Who's affected

- Anyone using Azure Storage Queues transport version 7.4.x and above

## Symptoms

When using ASQ with default connection string and alias to secure it, messages sent with raw connection string instead of alias.

### Details

Azure Storage Queues transport (ASQ) has a feature to [secure connections strings](https://docs.particular.net/transports/azure-storage-queues/multi-storageaccount-support#cross-namespace-routing-using-aliases-instead-of-connection-strings).

> In order to prevent accidentally leaking connection string values, it is recommended to use aliases instead of raw connection strings. When applied, raw connection string values are replaced with registered aliases removing the possibility of leaking a connection string value. **When using a single account, aliasing connection string is limited to just enabling it.** When multiple accounts are used, an alias has to be registered for each storage account.

This appear to be broken when a single (default) connection string is used. A separate [bug issue is raised ](https://github.com/Particular/NServiceBus.AzureStorageQueues/issues/308).
Unfortunately, the API is broken in a way that leads customers to assume that the default connection string is secured by invoking the API, when in reality it is not.

When API is called

```c#
transport.UseAccountAliasesInsteadOfConnectionStrings();
transport.DefaultAccountAlias("account_A");
```

Sending messages will contain the raw connection string rather than its alias as per documentation.

```json
{
  "IdForCorrelation": null,
  "Id": "7427f558-9e38-4959-9561-a869017ea425",
  "MessageIntent": 1,
  "ReplyToAddress": "samples-azure-storagequeues-endpoint1@UseDevelopmentStorage=true",
  "Headers": {
    "NServiceBus.MessageId": "7427f558-9e38-4959-9561-a869017ea425",
    "NServiceBus.MessageIntent": "Send",
    "NServiceBus.ReplyToAddress": "samples-azure-storagequeues-endpoint1@UseDevelopmentStorage=true",
  },
  "Body": "...",
  "CorrelationId": "7427f558-9e38-4959-9561-a869017ea425",
  "Recoverable": true
}
```

### Expected behaviour

ASQ should [behave as ASB](https://docs.particular.net/transports/azure-service-bus/securing-connection-strings#backward-compatibility) and provide a default alias, which is used for the default connection string when connection strings should be secured.

SeanFeldman:
[...]

PR #313, opened 2018-01-17
Hotfix-7.4.3

SeanFeldman:
Connects to #312

Fixes bug #308 - When using ASQ with default connection string and alias to secure it, messages sent with raw connection string instead of alias.

### DIFF
--- a/src/AcceptanceTests/Configuration/When_using_alias_instead_of_connection_string_for_default_account.cs
+++ b/src/AcceptanceTests/Configuration/When_using_alias_instead_of_connection_string_for_default_account.cs
@@ -3,6 +3,7 @@
     using System.IO;
     using System.Threading.Tasks;
     using AcceptanceTesting;
+    using Azure.Transports.WindowsAzureStorageQueues;
     using Azure.Transports.WindowsAzureStorageQueues.AcceptanceTests;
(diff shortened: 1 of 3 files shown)


SeanFeldman:

src/AcceptanceTests/Configuration/When_using_alias_instead_of_connection_string_for_default_account.cs
```diff
@@ -0,0 +1,77 @@
...
+    using Microsoft.WindowsAzure.Storage;
+    using Microsoft.WindowsAzure.Storage.Queue;
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Linq;
+    using NUnit.Framework;
+
+    public class When_using_alias_instead_of_connection_string_for_default_account : NServiceBusAcceptanceTest
+    {
+        CloudQueue destinationQueue;
+        const string EndpointName = nameof(SenderEndpoint);
```
danielmarbach: same comment as in other PR. Shouldn't this be using the endpoint naming convention?

SeanFeldman: See reply in the other PR.

PR #317, opened 2018-01-24
[Spike] Catch stop exception

timbussmann:
I noticed two things while testing a change to core to call start/stop on the transport infrastructure when running the transport tests and wanted to get some thoughts from the @Particular/azure-maintainers regarding those. This PR's purpose is mostly get a better understanding of the current code and should be closed once discussed.

* stop can throw a `TaskCanceledException` which won't be handled anywhere. It seems this exception is quite expected as it skips a call to `Task.Delay`. Core will catch this exception and log an error which seems weird as this doesn't really seem to be an actual exception? (Also, the transport tests will have to catch that exception too once we start calling Start/Stop on the transport infrastructure).

* `GetRequiredTransactionMode` in the `Start` method uses the settings to get access to the `TransportInfrastructure` class (`settings.Get<TransportInfrastructure>().TransactionMode`). This seems rather weird as this code is called from within the class implementing the `TransportInfrastructure`. Once the transport tests start calling Start on the transport infrastructure, this would fail as the test doesn't set this setting. You can argue that this doesn't correspond with core's behavior, I don't understand why this code is written that way though as it seems it could be easily avoided in the first place.

ping @Particular/azure-maintainers

### DIFF
--- a/src/Transport/AzureStorageQueueInfrastructure.cs
+++ b/src/Transport/AzureStorageQueueInfrastructure.cs
@@ -193,7 +193,7 @@
         {
             if (PollerCanBeUsed())
             {
-                var isAtMostOnce = GetRequiredTransactionMode(settings) == TransportTransactionMode.None;
+                var isAtMostOnce = GetRequiredTransactionMode() == TransportTransactionMode.None;
                 var maximumWaitTime = settings.Get<TimeSpan>(WellKnownConfigurationKeys.ReceiverMaximumWaitTimeWhenIdle);
                 var peekInterval = settings.Get<TimeSpan>(WellKnownConfigurationKeys.ReceiverPeekInterval);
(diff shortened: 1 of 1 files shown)


danielmarbach:
I agree @timbussmann there is no need to access the settings and the poller should safeguard on stop against the operation canceled exception.

SeanFeldman:
👍
I suggest we rebase this against `develop` and keep the changes. Any objections?

danielmarbach:
For me the graceful shutdown behavior belongs into the poller. Other than
that I'm OK

Am 27.01.2018 00:02 schrieb "Sean Feldman" <notifications@github.com>:

timbussmann:
I'm happy to split the PR and target the changes against develop if you're fine with the changes. Will follow up with that soonish.

danielmarbach:
@timbussmann sent in the following PR https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/319 containing your settings changes and the graceful shutdown in the poller. Are you OK with closing this spike PR?

PR #326, opened 2018-07-10
Delayed delivery polling fix

### LINKED ISSUE #324: DelayedMessagesPoller polls too often when no delayed messages are around
When enabling debug level logging, you'll notice that the `DelayedMessagesPoller` reports a polling attempt multiple times per second, even when no delayed message is around. Looking at the implementation, at first glance it doesn't look like the poller is supposed to behave that way and talking to @SeanFeldman confirmed that assumption further.

This has been brought up via support case from a customer noticing the polling attempts to his storage account. I've verified this behavior using the ASQ transport sample and changing the log level to debug.

ping @Particular/azure-maintainers

Scooletz:
Fixes #324 

This PR targets `develop`. Depending on what you want to do with it I might need to change the target branch as well as make it a hotfix

### DIFF
--- a/src/AcceptanceTests/Sending/When_delaying_messages_natively.cs
+++ b/src/AcceptanceTests/Sending/When_delaying_messages_natively.cs
@@ -1,7 +1,6 @@
 ﻿namespace NServiceBus.Azure.Transports.WindowsAzureStorageQueues.AcceptanceTests.Sending
 {
     using System;
-    using System.Collections.Concurrent;
     using System.Collections.Generic;
     using System.Diagnostics;
     using System.Linq;
--- a/src/AcceptanceTests/Sending/When_delaying_messages_natively.cs
+++ b/src/AcceptanceTests/Sending/When_delaying_messages_natively.cs
@@ -61,25 +61,36 @@
                 var delay = Task.Delay(TimeSpan.FromSeconds(15));
 
                 await Scenario.Define<Context>()
-                    .WithEndpoint<Sender>()
+                    .WithEndpoint<SlowlyPeekingSender>()
                     .Done(c => delay.IsCompleted)
                     .Run()
 
                     .ConfigureAwait(false);
 
                 Func<RequestEventArgs, Uri> map;
-#if NETCOREAPP2_0
+#if NETCOREAPP
 map = e => e.RequestUri;
 #else
                 map = e => e.Request.RequestUri;
 #endif
-                var dates = requests.Events.Where(e =>
-                {
-                    var lowered = map(e).ToString().ToLowerInvariant();
-                    return lowered.Contains(SenderDelayedMessagesTable.ToLowerInvariant()) && lowered.Contains("$filter");
-                }).Select(e => e.RequestInformation.StartTime).ToArray();
+                var requestCount = requests.Events
+                    .Where(e =>
+                    {
+                        var lowered = map(e).ToString().ToLowerInvariant();
+                        return lowered.Contains(SenderDelayedMessagesTable.ToLowerInvariant()) && lowered.Contains("$filter");
+                    })
+                    .Count();
 
-                Assert.Less(dates.Length, 15);
+                // the wait times for next peeks for SlowlyPeekingSender
+                // peek number  |wait (seconds)| cumulative wait (seconds)
+                // 1 | 0 | 0
+                // 2 | 1 | 1
+                // 3 | 2 | 3
+                // 4 | 3 | 6
+                // 5 | 4 | 10
+                // 6 | 5 | 15
+                // 7 | 6 | 21 <- this is the boundary
+                Assert.LessOrEqual(requestCount, 7);
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/Transport/AzureStorageQueueInfrastructure.cs
```diff
@@ -196,7 +196,7 @@ public override Task Start()
                 var isAtMostOnce = GetRequiredTransactionMode() == TransportTransactionMode.None;
                 var maximumWaitTime = settings.Get<TimeSpan>(WellKnownConfigurationKeys.ReceiverMaximumWaitTimeWhenIdle);
                 var peekInterval = settings.Get<TimeSpan>(WellKnownConfigurationKeys.ReceiverPeekInterval);
-                poller = new DelayedMessagesPoller(delayedDelivery.Table, connectionString, settings.ErrorQueueAddress(), isAtMostOnce, BuildDispatcher(), new BackoffStrategy(maximumWaitTime, peekInterval));
+                poller = new DelayedMessagesPoller(delayedDelivery.Table, connectionString, settings.ErrorQueueAddress(), isAtMostOnce, BuildDispatcher(), new BackoffStrategy(peekInterval, maximumWaitTime));
```
SeanFeldman: Awesome 😃

src/AcceptanceTests/Sending/When_delaying_messages_natively.cs
```diff
@@ -50,6 +53,36 @@ public class When_delaying_messages_natively : NServiceBusAcceptanceTest
...
+#else
+                map = e => e.Request.RequestUri;
+#endif
+                var dates = requests.Events.Where(e =>
+                {
+                    var lowered = map(e).ToString().ToLowerInvariant();
+                    return lowered.Contains(SenderDelayedMessagesTable.ToLowerInvariant()) && lowered.Contains("$filter");
+                }).Select(e => e.RequestInformation.StartTime).ToArray();
+
+                Assert.Less(dates.Length, 15);
```
SeanFeldman: Magic number... I would never understand why it's less than 15 and not 10. Would be good to add some description here.
 
| Iteration | Current Delay (mSec) | Total Delay (mSec) |
|-----------|------------------------|----------------------|
| 1 |                             250.0 |                        250.0 |
| 2 |                             375.0 |                        625.0 |
| 3 |                             500.0 |                     1,125.0 |
| 4 |                             625.0 |                     1,750.0 |
| 5 |                             750.0 |                     2,500.0 |
| 6 |                             875.0 |                     3,375.0 |
| 7 |                          1,000.0 |                     4,375.0 |
| 8 |                          1,125.0 |                     5,500.0 |
| 9 |                          1,250.0 |                     6,750.0 |
| 10 |                          1,375.0 |                     8,125.0 |
| 11 |                          1,500.0 |                     9,625.0 |
| 12 |                          1,625.0 |                   11,250.0 |
| 13 |                          1,750.0 |                   13,000.0 |
| **14** |                          1,875.0 |                   **14,875.0**  <-- within 15 seconds|
| 15 |                          2,000.0 |                   16,875.0 |

src/Transport/AzureStorageQueueInfrastructure.cs
Scooletz: Let's keep this private, ok? Oh man, this is a public PR 💥

src/AcceptanceTests/Sending/When_delaying_messages_natively.cs
Scooletz: Frankly speaking, I was thinking about providing this. We use Task.Delay though, which is executed after the specified period. There's a lot of asynchronicity in here. What I did to find this number I run this tests several times finding a number that lies more less in the middle of not failing for `back-off` and failing for `no-back-off`. What do you think?

src/AcceptanceTests/Sending/When_delaying_messages_natively.cs
```diff
@@ -50,6 +53,36 @@ public class When_delaying_messages_natively : NServiceBusAcceptanceTest
...
+
+                await Scenario.Define<Context>()
+                    .WithEndpoint<Sender>()
+                    .Done(c => delay.IsCompleted)
+                    .Run()
+
+                    .ConfigureAwait(false);
+
+                Func<RequestEventArgs, Uri> map;
+#if NETCOREAPP2_0
```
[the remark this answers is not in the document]
Scooletz: I can bet that there's going to be a breaking change in this API for 2.1 😛 but sure, let's do it 😉

src/AcceptanceTests/Sending/When_delaying_messages_natively.cs
Scooletz: Ok, I changed the peek time to 1s please my math-fu and provided the table

src/Tests/BackoffStrategyTests.cs
```diff
@@ -0,0 +1,45 @@
...
+            var strategy = new BackoffStrategy(peekInterval, maxWaitTime);
+
+            const int emptyBatchCount = 5;
+            for (var i = 0; i < emptyBatchCount; i++)
+            {
+                await strategy.OnBatch(0, CancellationToken.None);
+            }
+
+            var lowerBoundary = (int)(peekInterval.TotalMilliseconds * (emptyBatchCount - 1));
+            var delay = Task.Delay(lowerBoundary);
```
danielmarbach: The only thing I'm not really glad about it is that those tests are time-dependent. But I'm guessing investing more probably doesn't make much sense. The only way we could properly assert it would be by having the BackoffStrategy a virtual method that encapsulates the Task.Delay. Then have a testable implementation and assert the value passed to the overridden method.

Scooletz: I'm not happy with this as well. I treat it as 1st part of finding the error (the 2nd is the acceptance test that I provided). For a bug fix, imo, this is just enough.

PR #334, opened 2018-09-24
Update Storage client and extend default TTL

### LINKED ISSUE #325: Update Storage library to the latest (9.x) and enable infinite TTL
Version 9.x of Azure Storage library (2017-07-29 REST API) has added support for infinite TTL on queue messages.  This means the transport can support 

1. TTBR > 7 days
1. Store messages infinitely w/o the worry for loosing messages

Additionally, dependencies on OData libraries where removed, reducing footprint of this libarary.

Note: to enable infinite TTL, TTL parameter has to be specified.
```c#
await queue.AddMessageAsync(msg,timeToLive: TimeSpan.MaxValue, // infinite TTL or value of TTBR
null, new QueueRequestOptions(), new OperationContext());
```

image

SeanFeldman:
Fixes #325 

* New Storage client 9.3.2
* Default message TTL set to 30 days (not using infinite)

### DIFF
--- a/src/AcceptanceTests/NServiceBus.AzureStorageQueues.AcceptanceTests.csproj
+++ b/src/AcceptanceTests/NServiceBus.AzureStorageQueues.AcceptanceTests.csproj
@@ -11,7 +11,7 @@
 
   <!-- Force latest versions -->
   <ItemGroup>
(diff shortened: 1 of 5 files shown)


SeanFeldman:
Replaced with #334

PR #335, opened 2018-09-24
Update Storage client and extend default TTL

### LINKED ISSUE #325: Update Storage library to the latest (9.x) and enable infinite TTL

SeanFeldman:
Fixes #325 
Supersedes #334 

* New Storage client 9.3.2
* Default message TTL set to 30 days (not using infinite)

### DIFF
--- a/src/Transport/NServiceBus.AzureStorageQueues.csproj
+++ b/src/Transport/NServiceBus.AzureStorageQueues.csproj
@@ -14,7 +14,7 @@
 
   <ItemGroup>
     <!-- Minimum requried version for WindowsAzure.Storage is 8.2.0 -->
-    <PackageReference Include="WindowsAzure.Storage" Version="[8.2.0, 9.0.0)" />
+    <PackageReference Include="WindowsAzure.Storage" Version="[9.3.2, 10.0.0)" />
     <PackageReference Include="NServiceBus" Version="[7.0.0, 8.0.0)" />
     <PackageReference Include="Fody" Version="2.2.1" PrivateAssets="All" />
     <PackageReference Include="Janitor.Fody" Version="1.5.2" PrivateAssets="All" />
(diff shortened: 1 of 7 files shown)


src/Transport/NServiceBus.AzureStorageQueues.csproj
```diff
@@ -14,7 +14,7 @@
 
   <ItemGroup>
     <!-- Minimum requried version for WindowsAzure.Storage is 8.2.0 -->
-    <PackageReference Include="WindowsAzure.Storage" Version="[8.2.0, 9.0.0)" />
+    <PackageReference Include="WindowsAzure.Storage" Version="[9.3.2, 10.0.0)" />
```
danielmarbach: Is TTL only available in 9.3.2 and higher? Or why do we need this specific version

SeanFeldman: TTL is available from 9.0.0
We're using the latest version of the storage client. What would be the point to go back to 9.0.0 that was release [8 months ago](https://www.nuget.org/packages/WindowsAzure.Storage/9.0.0)?

danielmarbach: Well, this is a nuspec range. By setting the lowest to the latest version clients that are using this library are forced to upgrade to that version. So if there is no specific reason to require that specific lowest version it would be better to support a range.

yvesgoeleven: I agree, start from 9.0.0

SeanFeldman:
@yvesgoeleven please review changes.

PR #345, opened 2019-01-15
Use core 7.1 timeout migration mode

boblangley:
Updates ASQ transport to utilize the new Core 7.1 migration mode to migrate from Timeout manager delayed messages to native delayed delivery.


src/Transport/AzureStorageQueueInfrastructure.cs
```diff
@@ -25,6 +25,8 @@ internal AzureStorageQueueInfrastructure(SettingsHolder settings, string connect
             var timeoutManagerFeatureDisabled = settings.GetOrDefault<FeatureState>(typeof(TimeoutManager).FullName) == FeatureState.Disabled;
             var sendOnlyEndpoint = settings.GetOrDefault<bool>("Endpoint.SendOnly");
 
+            settings.SetDefault(WellKnownConfigurationKeys.DelayedDelivery.EnableMigrationMode, !timeoutManagerFeatureDisabled && !sendOnlyEndpoint);
```
danielmarbach: the send only check is not necessary I think, see https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutManager.cs#L20

src/Transport/AzureStorageQueueInfrastructure.cs
```diff
@@ -35,20 +37,7 @@ internal AzureStorageQueueInfrastructure(SettingsHolder settings, string connect
...
-                yield return typeof(NonDurableDelivery);
-
-                if (DelayedDeliveryCanBeUsed())
-                {
-                    yield return typeof(DoNotDeliverBefore);
-                    yield return typeof(DelayDeliveryWith);
-                }
-            }
-        }
+        public override IEnumerable<Type> DeliveryConstraints => new List<Type> {typeof(DiscardIfNotReceivedBefore), typeof(NonDurableDelivery), typeof(DoNotDeliverBefore), typeof(DelayDeliveryWith)};
```
danielmarbach: it doesn't really matter because it is not on the hot path but this property could have a regular setter and set the list once in the property initializer

src/Transport/AzureStorageQueueInfrastructure.cs
```diff
@@ -25,6 +25,8 @@ internal AzureStorageQueueInfrastructure(SettingsHolder settings, string connect
             var timeoutManagerFeatureDisabled = settings.GetOrDefault<FeatureState>(typeof(TimeoutManager).FullName) == FeatureState.Disabled;
```
danielmarbach: Could this use https://github.com/Particular/NServiceBus/blob/65fee422bdd79a3e5d4adc3dcc40353cb5b9dd82/src/NServiceBus.Core/Features/SettingsExtensions.cs#L45 ?

danielmarbach: See also https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/7e761293992ce8fd47acd561dda651bb22ebfb9e/src/Transport/DelayDelivery/NativeDelayDelivery.cs#L54

src/Transport/AzureStorageQueueInfrastructure.cs
danielmarbach: https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/7e761293992ce8fd47acd561dda651bb22ebfb9e/src/Transport/DelayDelivery/NativeDelayDelivery.cs#L55 could be using the migration mode flag?

src/Transport/AzureStorageQueueInfrastructure.cs
bording: I think the existing approach makes more sense because core doesn't call this often (maybe once?), so this lets the collection be garbage collected instead of keeping unneeded memory allocated for the lifetime of the application.

src/Transport/AzureStorageQueueInfrastructure.cs
bording: The check for SendOnly is an optimization only to prevent users from needing to call the `DisableTimeoutManager` API when they've already called an API to make the endpoint send only.

src/Transport/AzureStorageQueueInfrastructure.cs
bording: I suppose it could, but I'm also not sure it's worth it?

danielmarbach: I don't see the argument why using an already existing API by the core is not worth it.

src/Transport/AzureStorageQueueInfrastructure.cs
danielmarbach: I'm ok with that

src/Transport/AzureStorageQueueInfrastructure.cs
bording: Well, I only meant that the code in question wasn't new to the PR, it was already there, so would it be worth changing now or not.

danielmarbach: And I said I'm fine either way ;)

PR #396, opened 2020-02-25
Improve log levels

### LINKED ISSUE #101: Improve log levels
Trying to identify why #86 issue happened, log was useless:
- when we don't have access to source code (nuget package referenced or working on customer code) understanding what's happening isn't easy. `Debug` level should log each method has been invoked, with parameters.
- here https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/develop/src/Transport/MessagePump.cs#L199 log level is wrong (IMO): that's an error. Review all log levels.

Connects to https://github.com/Particular/NServiceBus.Azure/issues/309

WilliamBZA:
Closes #101 

Adds a bunch of log debug messages. I've tried to include message IDs wherever I can see that it makes sense.

### DIFF
--- a/src/Transport/BackoffStrategy.cs
+++ b/src/Transport/BackoffStrategy.cs
@@ -26,7 +26,7 @@ namespace NServiceBus.Transport.AzureStorageQueues
 
         void OnSomethingProcessed()
         {
-            Logger.Debug("Processed message, setting delay until next peek to 0");
+            Logger.Debug("Processed message, setting delay of the next peek to 0 seconds");
             timeToDelayUntilNextPeek = TimeSpan.Zero;
         }
 
--- a/src/Transport/BackoffStrategy.cs
+++ b/src/Transport/BackoffStrategy.cs
@@ -1,5 +1,6 @@
 namespace NServiceBus.Transport.AzureStorageQueues
 {
+    using NServiceBus.Logging;
     using System;
     using System.Threading;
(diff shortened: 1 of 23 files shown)


src/Transport/MessageRetrieved.cs
```diff
@@ -28,6 +29,7 @@ public async Task<MessageWrapper> Unwrap()
         {
             try
             {
+                Logger.DebugFormat("Unwrappinbg message ID: '{0}'", rawMessage.Id);
```
indualagarsamy: typo : unwrapping

src/Transport/BackoffStrategy.cs
```diff
@@ -23,11 +26,14 @@ public BackoffStrategy(TimeSpan peekInterval, TimeSpan maximumWaitTimeWhenIdle)
 
         void OnSomethingProcessed()
         {
+            Logger.Debug("Processed message, setting delay until next peek to 0");
```
indualagarsamy: this message isn't very clear. Did you mean: Setting delay of next peek to 0 seconds?

src/Transport/MessageRetrieved.cs
SeanFeldman: Fixed

src/Transport/BackoffStrategy.cs
SeanFeldman: Fixed

PR #407, opened 2020-04-23
Move poisonous messages to the error queue

### LINKED ISSUE #405: Incoming messages failing unwrapping are processed infinitely, filling up the error queue.
### Symptoms

Incoming messages failing unwrapping are processed infinitely, filling up the error queue. The original incoming messages are not removed from the input queue and the following warning is logged:

```
WARN : Azure Storage Queue transport failed pushing a message through pipeline
Type: Microsoft.WindowsAzure.Storage.StorageException
Message: The remote server returned an error: (404) Not Found.
Source: Microsoft.WindowsAzure.Storage
StackTrace:
   at Microsoft.WindowsAzure.Storage.Core.Executor.Executor.EndExecuteAsync[T](IAsyncResult result) in c:\Program Files (x86)\Jenkins\workspace\release_dotnet_master\Lib\ClassLibraryCommon\Core\Executor\Executor.cs:line 50
   at Microsoft.WindowsAzure.Storage.Core.Util.AsyncExtensions.<>c__DisplayClass7.<CreateCallbackVoid>b__5(IAsyncResult ar) in c:\Program Files (x86)\Jenkins\workspace\release_dotnet_master\Lib\ClassLibraryCommon\Core\Util\AsyncExtensions.cs:line 121
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at NServiceBus.Transport.AzureStorageQueues.MessageRetrieved.<Unwrap>d__3.MoveNext() in C:\BuildAgent\work\3c19e2a032c05076\src\Transport\MessageRetrieved.cs:line 40
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at NServiceBus.Transport.AzureStorageQueues.MessagePump.<InnerReceive>d__7.MoveNext() in C:\BuildAgent\work\3c19e2a032c05076\src\Transport\MessagePump.cs:line 153
TargetSite: T EndExecuteAsync[T](System.IAsyncResult)
```

### Who's affected

Customers using Azure Storage Queue transport with a [custom envelope unwrapper](https://docs.particular.net/transports/azure-storage-queues/configuration#custom-envelope-unwrapper).

### Root cause

When a message arrives, it's handled as a `MessageRetrieved`.  The first step before a message can be processed is to unwrap it. If the message is poisonous or a custom unwrapper is failing, the message is supposed to be moved to the error queue. Instead, the message will remain in the main queue, will cause infinite processing, spamming error queue with infinite copies of the message.

The root cause of the issue is the implementation of the `MessageRetrieved.Unrawp()` method that is sending the originally retrieve message to the error queue and complete it using the same object. The Storage Queue API when sending a message updated message ID and its pop receipt. Therefore the original message's information is lost and it cannot be completed. From the [documentation](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.storage.queue.cloudqueue.addmessageasync?view=azure-dotnet-legacy#Microsoft_Azure_Storage_Queue_CloudQueue_AddMessageAsync_Microsoft_Azure_Storage_Queue_CloudQueueMessage_):

> The CloudQueueMessage message passed in will be populated with the pop receipt, message ID, and the insertion/expiration time.

Originally reported on [StackOverflow](https://stackoverflow.com/questions/61329547/nservicebus-with-azurestoragequeues-not-removing-poison-messages-from-input-queu).

SeanFeldman:
Fixes #405

### DIFF
--- a/src/Transport/MessageRetrieved.cs
+++ b/src/Transport/MessageRetrieved.cs
@@ -34,6 +34,9 @@
             }
             catch (Exception ex)
             {
+                // When a CloudQueueMessage retrieved from a is en-queued directly, message's ID and PopReceipt are mutated.
+                // To be able to delete the original message, original message ID and PopReceipt have to be stored aside.
+
                 var messageId = rawMessage.Id;
                 var messagePopReceipt = rawMessage.PopReceipt;
 
--- a/src/Transport/MessageRetrieved.cs
+++ b/src/Transport/MessageRetrieved.cs
@@ -34,10 +34,13 @@
             }
             catch (Exception ex)
             {
-                await errorQueue.AddMessageAsync(rawMessage).ConfigureAwait(false);
-                await inputQueue.DeleteMessageAsync(rawMessage).ConfigureAwait(false);
+                var messageId = rawMessage.Id;
+                var messagePopReceipt = rawMessage.PopReceipt;
 
-                throw new SerializationException($"Failed to deserialize message envelope for message with id {rawMessage.Id}. Make sure the configured serializer is used across all endpoints or configure the message wrapper serializer for this endpoint using the `SerializeMessageWrapperWith` extension on the transport configuration. Please refer to the Azure Storage Queue Transport configuration documentation for more details.", ex);
+                await errorQueue.AddMessageAsync(rawMessage).ConfigureAwait(false);
+                await inputQueue.DeleteMessageAsync(messageId, messagePopReceipt).ConfigureAwait(false);
+
+                throw new SerializationException($"Failed to deserialize message envelope for message with id {messageId}. Make sure the configured serializer is used across all endpoints or configure the message wrapper serializer for this endpoint using the `SerializeMessageWrapperWith` extension on the transport configuration. Please refer to the Azure Storage Queue Transport configuration documentation for more details.", ex);
             }
         }
 
--- a/src/Transport/MessageRetrieved.cs
+++ b/src/Transport/MessageRetrieved.cs
@@ -34,7 +34,7 @@
             }
             catch (Exception ex)
             {
-                // When a CloudQueueMessage retrieved from a is en-queued directly, message's ID and PopReceipt are mutated.
+                // When a CloudQueueMessage is retrieved and is en-queued directly, message's ID and PopReceipt are mutated.
                 // To be able to delete the original message, original message ID and PopReceipt have to be stored aside.
 
                 var messageId = rawMessage.Id;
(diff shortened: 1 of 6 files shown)


src/Transport/MessageRetrieved.cs
```diff
@@ -34,10 +34,13 @@ public async Task<MessageWrapper> Unwrap()
             }
             catch (Exception ex)
             {
+                var messageId = rawMessage.Id;
```
danielmarbach, line 37: Would it also make sense to use the same mechanism in the Ack method?

SeanFeldman: Here we're forced to do so due to, arguably, unclear behaviour of the `.AddMessageAsync(message)` method that mutates the passed in message and overriding the original values for ID and pop receipt. 

The SDK is leveraging message ID and pop receipt from the message passed in. By using the overload taking ID and pop receipt we'll not gain any substantial benefits over passing a message. We will increase the code though. I would suggest keeping the `Ack` method as-is.

danielmarbach: Ok fair enough. I was just worried we might not remember why we use two different mechanisms.

SeanFeldman: Valid concern. I don't mind to add a comment here to clarify the intent. Would that be helpful?

SeanFeldman: Added 2573831525381d5fb69dd947f9f502e40cc0b5b1

danielmarbach: Perfect

src/Transport/MessageRetrieved.cs
```diff
@@ -34,10 +34,16 @@ public async Task<MessageWrapper> Unwrap()
             }
             catch (Exception ex)
             {
+                // When a CloudQueueMessage retrieved from a is en-queued directly, message's ID and PopReceipt are mutated.
```
danielmarbach, line 37: Sorry can't make a suggestion because I'm on mobile. Looks like this sentence is incomplete. Should this be "from a queue..."?

SeanFeldman: Not quite but it was a broken sentence. Fixed it.

PR #427, opened 2020-07-08
Change log level for delayed to info

WilliamBZA:
Logs an info when delayed messages are as a result of contention

### DIFF
--- a/src/Transport/DelayDelivery/DelayedMessagesPoller.cs
+++ b/src/Transport/DelayDelivery/DelayedMessagesPoller.cs
@@ -67,6 +67,10 @@
                 {
                     // ok, since the InnerPoll could observe the token
                 }
+                catch (StorageException ex) when (ex.RequestInformation.HttpStatusCode == 409)
+                {
+                    Logger.Info("Failed to fetch delayed messages from the storage", ex);
+                }
                 catch (Exception ex)
                 {
                     Logger.Warn("Failed to fetch delayed messages from the storage", ex);


src/Transport/DelayDelivery/DelayedMessagesPoller.cs
```diff
@@ -67,6 +67,10 @@ await InnerPoll(cancellationToken)
                 {
                     // ok, since the InnerPoll could observe the token
                 }
+                catch (StorageException ex) when (string.Equals(ex.Message, "The remote server returned an error: (409) Conflict", StringComparison.InvariantCultureIgnoreCase))
+                {
+                    Logger.Info("Failed to fetch delayed messages from the storage", ex);
+                }
                 catch (Exception ex)
                 {
                     Logger.Warn("Failed to fetch delayed messages from the storage", ex);
```
SeanFeldman, line 76: Should probably do the same here

SeanFeldman: Nope, we shouldn't. If there's no conflict, it's an exception we should warn about.

WilliamBZA: I added the filter and specific exception type because I wasn't sure if all exceptions should be info, if they should then I'll remove as less code == better

SeanFeldman: 409 is going to happen more often just because scaled-out instances will be competing over the entries.
Is there a way to filter on the status code rather than string?

src/Transport/DelayDelivery/DelayedMessagesPoller.cs
```diff
@@ -67,6 +67,10 @@ await InnerPoll(cancellationToken)
                 {
                     // ok, since the InnerPoll could observe the token
                 }
+                catch (StorageException ex) when (ex.RequestInformation.HttpStatusCode == 409)
```
SeanFeldman, line 70: ❤️

danielmarbach:
I sent in an alternative after trying to help @WilliamBZA to get to the bottom of this issue

https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/428

The PR targets master. After the problem is fixed on master it can be backported to 7.5. I'll propose to close this one

PR #429, opened 2020-07-09
Only allocate stopwatch when there is something to dispatch

danielmarbach:


src/Transport/DelayDelivery/DelayedMessagesPoller.cs
```diff
@@ -150,15 +150,20 @@ async Task SpinOnce(CancellationToken cancellationToken)
...
                 if (cancellationToken.IsCancellationRequested)
                 {
                     return;
                 }
 
+                // only allocate if needed
+                if (stopwatch == null)
+                {
+                    stopwatch = Stopwatch.StartNew();
+                }
```
SeanFeldman, line 165: I'd suggest `stopwatch = stopwatch ?? Stopwatch.StartNew();` but don't think we can do it with c# language constrained to version 7.3.

How about this:

```suggestion
                stopwatch = stopwatch ?? Stopwatch.StartNew();
```

danielmarbach:
Like it Sean. Used your suggestion

PR #446, opened 2020-08-31
Azure Storage SDK update

SeanFeldman:
Spike PR to upgrade from `WindowsAzure.Storage` to `Azure.Storage.*`

Compiles, 1 test fails, smoke tested with the previous version interoperability

### DIFF
--- a/src/Transport/CreateQueueClients.cs
+++ b/src/Transport/CreateQueueClients.cs
@@ -1,31 +1,25 @@
 namespace NServiceBus.Transport.AzureStorageQueues
 {
     using System.Collections.Concurrent;
-    using Microsoft.WindowsAzure.Storage;
-    using Microsoft.WindowsAzure.Storage.Queue;
+    using global::Azure.Storage.Queues;
 
     class CreateQueueClients
     {
-        public CloudQueueClient Create(ConnectionString connectionString)
+        public QueueServiceClient Create(ConnectionString connectionString)
         {
             return destinationQueueClients.GetOrAdd(connectionString, cs => BuildClient(cs));
         }
 
-        public static CloudQueueClient CreateReceiver(ConnectionString connectionString)
+        public static QueueServiceClient CreateReceiver(ConnectionString connectionString)
         {
             return BuildClient(connectionString);
         }
 
-        static CloudQueueClient BuildClient(ConnectionString connectionString)
+        static QueueServiceClient BuildClient(ConnectionString connectionString)
         {
-            if (CloudStorageAccount.TryParse(connectionString.Value, out var account))
-            {
-                return account.CreateCloudQueueClient();
-            }
-
-            return null;
+            return new QueueServiceClient(connectionString.Value);
         }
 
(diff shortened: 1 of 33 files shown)


SeanFeldman:
Smoked tested with one of the [ASQ samples](https://docs.particular.net/samples/azure/storage-queues/). Good to go.

src/Transport/CreateQueueClients.cs
```diff
@@ -1,31 +1,25 @@
...
 {
     using System.Collections.Concurrent;
-    using Microsoft.WindowsAzure.Storage;
-    using Microsoft.WindowsAzure.Storage.Queue;
+    using global::Azure.Storage.Queues;
 
     class CreateQueueClients
     {
-        public CloudQueueClient Create(ConnectionString connectionString)
+        public QueueServiceClient Create(ConnectionString connectionString)
```
danielmarbach, line 8: It seems that previously we have been able to cache the QueueClients directly but now we cache the service client and have to acquire a new queue client for every dispatch. Judging by the code the `GetQueueClient` creates a new `QueueClient` for every call. So essentially we have now a queue client object per dispatch. I see it is possible to create a QueueClient directly but then we would have to also know the queue name in this class and store them by combination of connection string as well as queue client. Are there any best practices we can leverage?

src/Transport/DefaultMessageEnvelopeUnwrapper.cs
```diff
@@ -13,11 +14,14 @@ public DefaultMessageEnvelopeUnwrapper(MessageWrapperSerializer messageSerialize
...
+        public MessageWrapper Unwrap(QueueMessage rawMessage)
         {
-            Logger.DebugFormat("Unwrapping native message (native ID: '{0}')", rawMessage.Id);
+            Logger.DebugFormat("Unwrapping native message (native ID: '{0}')", rawMessage.MessageId);
             MessageWrapper m;
-            using (var stream = new MemoryStream(rawMessage.AsBytes))
+
+            var bytes = Convert.FromBase64String(rawMessage.MessageText);
+
+            using (var stream = new MemoryStream(bytes))
```
danielmarbach, line 24: This could also benefit from recyclable memory stream later

src/Transport/Dispatcher.cs
```diff
@@ -62,7 +62,7 @@ public async Task Send(UnicastTransportOperation operation, CancellationToken ca
 
             var sendClient = createQueueClients.Create(connectionString);
             var q = addressGenerator.GetQueueName(queue.QueueName);
-            var sendQueue = sendClient.GetQueueReference(q);
+            var sendQueue = sendClient.GetQueueClient(q);
```
danielmarbach, line 65: See https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/446/files#r508241639

src/Transport/TaskEx.cs
```diff
@@ -9,8 +9,5 @@ static class TaskEx
         public static void Ignore(this Task task)
         {
         }
-
-        //TODO: remove when we update to 4.6 and can use Task.CompletedTask
-        public static readonly Task CompletedTask = Task.FromResult(0);
```
danielmarbach, line 14: 👍

danielmarbach (approved):

PR #468, opened 2020-10-29
Breaking changes related to removal of timeout manager

ramonsmits:
- Fix breaking changes related to removal of timeout manager
- Marked code related to non-durable messaging with TODO's
- Fix change related to AT persistence configuration

### DIFF
--- a/src/AcceptanceTests/Sending/When_delaying_messages_and_delayed_delivery_is_disabled.cs
+++ b/src/AcceptanceTests/Sending/When_delaying_messages_and_delayed_delivery_is_disabled.cs
@@ -13,7 +13,7 @@
         public void Should_throw()
         {
             Context context = null;
-            var exception = Assert.ThrowsAsync<Exception>(async () => await Scenario.Define<Context>(ctx => context = ctx)
+            var exception = Assert.ThrowsAsync<InvalidOperationException>(async () => await Scenario.Define<Context>(ctx => context = ctx)
                     .WithEndpoint<Endpoint>(b => b.When((session, ctx) =>
                     {
                         var delay = TimeSpan.FromSeconds(2);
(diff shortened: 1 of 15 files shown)


src/Transport/obsoletes.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace NServiceBus
+{
+    using System;
+
+    /// <summary>Configures native delayed delivery.</summary>
```
bording, line 5: XML comments should be removed from obsoletes. The file can be marked with `#pragma warning disable 1591` / `#pragma warning restore 1591`

src/Transport/Config/DelayedDeliverySettings.cs
```diff
@@ -46,9 +37,6 @@ public void DisableDelayedDelivery()
         {
             // disable delayed delivery
             this.GetSettings().Set(WellKnownConfigurationKeys.DelayedDelivery.DisableDelayedDelivery, true);
```
bording, line 39: This setting and the test about throwing when it's disable means there is more work to done here. It's not the transports responsibility to throw in this scenario now. Instead, we need to alter the `DeliveryConstraints` in  `AzureStorageQueueInfrastructure` when this is called to inform core and let it handle things.

src/AcceptanceTests/Sending/When_delaying_messages_and_delayed_delivery_is_disabled.cs
```diff
@@ -46,7 +46,6 @@ public Endpoint()
                 EndpointSetup<DefaultServer>(config =>
                 {
                     var delayedDelivery = config.UseTransport<AzureStorageQueueTransport>().DelayedDelivery();
-                    delayedDelivery.DisableTimeoutManager();
```
bording, line 49: I suspect this entire test can be removed once we fix up the `DisableDelayedDelivery` API.

bording (changes requested):

src/AcceptanceTests/Sending/When_delaying_messages_and_delayed_delivery_is_disabled.cs
ramonsmits: @bording what do you mean, that there will be a test part of the core AT's to replace this?

bording (approved):

PR #470, opened 2020-11-02
Concurrent dispatch in delayed message poller

### LINKED ISSUE #430: Dispatching of due timeouts is done sequentially instead of concurrently
The delayed message poller fetches 50 entries at a time if available and then dispatches those entries sequentially. By concurrently dispatching all the timeouts we could greatly increase the timeout throughput and decrease the chance of running into lease expiry.

danielmarbach:
Fixes #430

### DIFF
--- a/src/Transport/DelayDelivery/DelayedMessagesPoller.cs
+++ b/src/Transport/DelayDelivery/DelayedMessagesPoller.cs
@@ -179,6 +179,7 @@
                     stopwatch.Reset();
                 }
 
+                // deliberately not using the dispatchers batching capability because every delayed message dispatch should be independent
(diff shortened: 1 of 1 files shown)


danielmarbach:
Previously the code did check

```
                // after half check if the lease is active
                if (stopwatch.Elapsed > HalfOfLeaseLength)
                {
                    if (await lockManager.TryLockOrRenew(cancellationToken).ConfigureAwait(false) == false)
                    {
                        return;
                    }
                    stopwatch.Reset();
                }
```

after every dispatch. Now the code only checks this before firing a concurrent task. I decided against making this more sophisticated since concurrent dispatch already makes this way less likely to even hit the lock renewal half lease time.

SeanFeldman (approved):

PR #473, opened 2020-11-03
Sanitization change

SeanFeldman:
Obsolete SanitizeQueueNamesWith and introduce an internal configuration for the test projects


src/Transport/obsoletes-v9.cs
```diff
@@ -24,5 +21,18 @@ partial class DelayedDeliverySettings
             RemoveInVersion = "10")]
         public void DisableTimeoutManager() => throw new NotImplementedException();
     }
+
+    public static partial class AzureStorageTransportExtensions
+    {
+        /// <summary>
```
danielmarbach, line 27: The comments can be removed. We disable 1591 in this file

danielmarbach (approved):

PR #474, opened 2020-11-04
Core AT tests already have tests for delayed delivery.

ramonsmits:


andreasohlund (approved):

danielmarbach (changes requested):
I'm 👎 to merge this. These tests verify the transport-specific behavior against the tables. These tests are crucial to cover the infrastructure visible side-effects from the perspective of the transport. 

I think we cannot assume the core test suite is always enough to come to the conclusion transport/infrastructure specific tests are no longer necessary. 

I think at least

- When_delaying_messages_natively_with_send_only 
- When_delaying_messages_natively 

should stay. It makes sense though to remove the hybrid test in my view since that behavior is no longer required right?

danielmarbach:
I think at least

- When_delaying_messages_natively_with_send_only 
- When_delaying_messages_natively 

should stay. It makes sense though to remove the hybrid test in my view since that behavior is no longer required right?

ramonsmits:
@danielmarbach You are right, I see 2 tests can be removed. The naming of the fixture seem to imply these could be safely removed but cannot.

ramonsmits:
@danielmarbach I think I've addressed your concerns. Glad that you chimed in :-)

danielmarbach (approved):

PR #479, opened 2020-11-06
cleanups

SeanFeldman:


src/Transport/Config/QueueAddress.cs
```diff
@@ -3,7 +3,7 @@
     using System;
     using Microsoft.Azure.Cosmos.Table;
 
-    readonly struct QueueAddress : IEquatable<QueueAddress>
```
danielmarbach, line 6: Why is this removed?

danielmarbach (approved):

src/Transport/Config/QueueAddress.cs
SeanFeldman: R# 🙂

danielmarbach: Yeah that is a false positive

PR #490, opened 2020-11-11
Align transport and package naming convention

danielmarbach:


src/Transport/AzureStorageQueueInfrastructure.cs
```diff
@@ -172,11 +172,9 @@ AzureStorageAddressingSettings GetAddressing(ReadOnlySettings settings, IProvide
                 accounts = new AccountConfigurations();
             }
 
-            const string defaultAccountAlias = "";
-
             addressing.SetAddressGenerator(addressGenerator);
-            addressing.RegisterMapping(accounts.defaultAlias ?? defaultAccountAlias, accounts.mappings);
-            addressing.Add(new AccountInfo(defaultAccountAlias, queueServiceClientProvider.Client), false);
+            addressing.RegisterMapping(accounts.defaultAlias, accounts.mappings);
```
SeanFeldman, line 176: Are you sure about this one?

SeanFeldman (approved):
Approving but that default alias... I saw that null causing an issue when exchanging message with an older endpoint.

danielmarbach:
Yeah that slipped through. Already fixed

danielmarbach:
`MessageWrapper` has still `NServiceBus.Azure.Transports.WindowsAzureStorageQueues` because I wasn't sure if it can be moved. Did we ever serialize the type info with ASQ? We probably didn't right? If that is true we could technically move it

src/Transport/AzureStorageQueueInfrastructure.cs
danielmarbach: Already force pushed and updated

PR #492, opened 2020-11-13
Connection string handling

SeanFeldman:
The connection string is becoming optional with the introduction of clients. Since it cannot be deprecated, we still need to support it yet throw when an attempt to use a Cosmos DB Table API connection string is used for either queue or blob client.

### DIFF
--- a/src/Tests/When_ConnectionStringValidator_is_used_with_cosmosdb_connection_string.cs
+++ b/src/Tests/When_ConnectionStringValidator_is_used_with_cosmosdb_connection_string.cs
@@ -1,19 +1,17 @@
 ﻿using System;
-using NServiceBus.Settings;
-using NServiceBus.Transport.AzureStorageQueues;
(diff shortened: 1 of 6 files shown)


danielmarbach (approved):

src/Tests/When_ConnectionStringValidator_is_used_with_cosmosdb_connection_string.cs
```diff
@@ -1,19 +1,17 @@
 ﻿using System;
-using NServiceBus.Settings;
-using NServiceBus.Transport.AzureStorageQueues;
```
danielmarbach, line 3: On the release branch these usings have been moved into the namespace

SeanFeldman: Yep.

PR #498, opened 2020-11-25
Sanitization should not be internalized

SeanFeldman:
Sanitization was internalized when should not have been


src/AcceptanceTests/ConfigureAzureStorageQueueTransport.cs
```diff
@@ -38,7 +38,7 @@ public Task Configure(string endpointName, EndpointConfiguration configuration,
         {
             Assert.Ignore("Ignored until issue #173 is resolved.");
         }
-
+
```
danielmarbach, line 41: 🔥

danielmarbach (approved):

src/AcceptanceTests/ConfigureAzureStorageQueueTransport.cs
SeanFeldman, line 41:
```suggestion

```

danielmarbach: The newline was correct but the space not 😉 anyway not a big deal. Feel free to move forward (that's why I approved)

PR #504, opened 2021-01-12
GitHubSync update - master

particularbot6:
This is an automated synchronization PR.

The following source template repositories were used:
* particular/RepoStandards


src/Transport/DelayDelivery/SimpleJSON.cs
```diff
@@ -62,14 +64,12 @@ namespace SimpleJson
...
     using System.Collections.Generic;
     using System.ComponentModel;
-    using System.Diagnostics.CodeAnalysis;
 
     /// <summary>
     /// Represents the json array.
     /// </summary>
     [GeneratedCode("simple-json", "1.0.0")]
     [EditorBrowsable(EditorBrowsableState.Never)]
-    [SuppressMessage("Microsoft.Naming", "CA1710:IdentifiersShouldHaveCorrectSuffix")]
```
bording, line 72: Looking at the [SimpleJson source](https://github.com/facebook-csharp-sdk/simple-json/blob/master/src/SimpleJson/SimpleJson.cs), all of these attributes are in there, and not something added by us, so I'm thinking when might be able to just revert all the changes to this file.

bording (changes requested):

src/Transport/DelayDelivery/SimpleJSON.cs
SeanFeldman: If we revert the change then we'll have errors about unnecessary suppressions.

bording: I didn't run into any errors with the copy of SimpleJson in Core, so went to take a look why that is. 

In Core, we've named the file `SimpleJson.g.cs`, so I think that is causing it to be skipped for rules.

Might be worth making that change instead of changing the contents?

SeanFeldman: Oh, that would work better. Will do.

bording (approved):

PR #539, opened 2021-03-03
Native publish/subscribe support on top of Azure Table Storage

danielmarbach:
The design is very similar to how the SQL transport native pubsub works. The TLDR is:

- Subscribers subscribe on the subscribed event type directly
- Polymorphism is implemented on the publisher side by disecting the published typed into its hierarchy and then reading all subscribers on those topics into a distinct list of queue addresses. By doing so automatic deduplication on the hierarchy can be guaranteed.

asq topologies implementation

All multi-cast operations are transformed into UnicastOperations to make sure multi-storage account support still works. This design also further reduces the complexity of the dispatcher.

### Multi storage account support

When the transport was still doing message-drive pubsub the multi-storage account support was relying on the publisher information and on the fact the subscription messages are sent to the publisher queue. With the native pubsub implementation, this approach no longer works. Instead, we have chosen a design that leverages the account alias metadata to derive the connection to the publisher's storage account and insert a subscription into the subscription table on the other storage account with a queue address using the subscriber alias.

Since publishers use unicast operations to directly send events into the subscriber queues, they have always required the subscriber alias and the connection to be configured on the publisher side. When the publisher dispatches the unicast operations it queries the interested subscribers which then returns local non-aliased, as well as remote and hence aliased, subscribers.

asq topologies multi-account


src/Transport/PubSub/SubscriptionStore.cs
```diff
@@ -15,9 +17,39 @@ public SubscriptionStore(CloudTable subscriptionTable)
...
-            return Task.FromResult(new List<string> { eventType.FullName });
+            var topics = GetTopics(eventType);
+
+            if (topics.Length == 0)
+            {
+                return Enumerable.Empty<string>();
+            }
+
+            var retrieveTasks = new List<Task<SubscriptionEntity>>(topics.Length);
+            var addresses = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```
yvesgoeleven, line 30: This is used for dedupe I assume?

src/Transport/PubSub/SubscriptionEntity.cs
```diff
@@ -0,0 +1,21 @@
...
+            set => PartitionKey = value;
+        }
+
+        public string Endpoint
+        {
+            get => RowKey;
+            set => RowKey = value;
+        }
+
+        public string Address { get; set; }
```
yvesgoeleven, line 19: Does this magically convert into a column under the hood?

src/Transport/PubSub/SubscriptionStore.cs
danielmarbach: yes. btw this is still WIP

src/Transport/PubSub/SubscriptionEntity.cs
danielmarbach: Yes

src/Transport/Config/SubscriptionSettings.cs
```diff
@@ -0,0 +1,67 @@
...
+
+
+        /// <summary>
+        ///     Default to 5 seconds caching. If a system is under load that prevent doing an extra roundtrip for each Publish
+        ///     operation. If
+        ///     a system is not under load, doing an extra roundtrip every 5 seconds is not a problem and 5 seconds is small enough
+        ///     value that
+        ///     people accepts as we always say that subscription operation is not instantaneous.
+        /// </summary>
+        TimeSpan cacheInvalidationPeriod = TimeSpan.FromSeconds(5);
```
SeanFeldman, line 61: Fills a bit aggressive but I have no idea what a normal load for ASQ customers would look like.

src/Transport/PubSub/CachedSubscriptionStore.cs
```diff
@@ -0,0 +1,60 @@
+namespace NServiceBus.Transport.AzureStorageQueues
+{
+    using System;
+    using System.Collections.Concurrent;
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+
+    class CachedSubscriptionStore : ISubscriptionStore
```
SeanFeldman, line 10: Nice use of the decorator pattern 👍

SeanFeldman (approved):

src/Transport/Config/SubscriptionSettings.cs
danielmarbach: I took the same value as SQL transport for now. I think it is a reasonable tradeoff as a default value. If you are publishing every 5 seconds then having some additional storage requests is not going to hurt you much but if you are publishing with greater cadence the benefits kicks in. I find this a sensible value also because caching is by default enabled. Users can always tweak it.

src/Transport/PubSub/CachedSubscriptionStore.cs
danielmarbach: Shamelessly stolen from SQL

yvesgoeleven (approved):

PR #554, opened 2021-03-12
remove or bump treat as error from version

yvesgoeleven:
@mauroservienti @timbussmann 

The PR is related to backporting the features in https://github.com/Particular/DeveloperExperience/issues/1249 to a new release branch, which will become v10.

As soon as we would create that branch gitversion will bump master to v11. That would cause all obsolete TreatAsErrorFromVersion checks that have been introduced by the new transport seam work to trigger and break the build.  

To mitigate these errors I have prepared this pr which either removes certain obsoletes or bumps them up.

Could you validate that I'm not making any mistakes in respect to the transport seam work.


timbussmann:
> The PR is related to backporting the features in Particular/DeveloperExperience#1249 to a new release branch, which will become v10.

I'm struggling to understand what you're planning to do.
How does backporting features create a v10 release branch? Are you planning to release a new major version of the transport targeting vore v7?

danielmarbach:
> Are you planning to release a new major version of the transport targeting vore v7?

@timbussmann correct. We have to because we are introducing native pubsub

timbussmann:
> @timbussmann correct. We have to because we are introducing native pubsub

🤔 you'll need soo much more to get master back on a core v7 state, at this point the only way that would be reasonable would be force pushing master to a previous state, the obsolete are literally the smallest problem?

yvesgoeleven:
@timbussmann master does not need to go back to core v7, it can stay on core v8... we will create an asq v10 release branch from v9 release branch. Because of this action the master build would break though as it will automatically jump to v11 due to how gitversion works, the fixes included in this pr will ensure that build will continue to work.

timbussmann:
> @timbussmann master does not need to go back to core v7, it can stay on core v8... we will create an asq v10 release branch from v9 release branch. Because of this action the master build would break though as it will automatically jump to v11 due to how gitversion works, the fixes included in this pr will ensure that build will continue to work.

gotcha, thanks. I'd still strongly prefer to not "backport" this in the first place but that's up to the TF & squad of course.

timbussmann (approved):

PR #555, opened 2021-03-12
Support native publish and subscribe with storage table

danielmarbach:
Backport of #547


src/AcceptanceTests/PubSub/When_migrating_publisher_first.cs
```diff
@@ -0,0 +1,190 @@
...
+                        c.GetSettings().Set("Transport.AzureStorageQueue.DisablePublishSubscribe", true);
+                    });
+                    b.When(c => c.SubscribedMessageDriven, (session, ctx) => session.Publish(new MyEvent()));
+                })
+
+                .WithEndpoint<Subscriber>(b =>
+                {
+                    b.CustomConfig(c =>
+                    {
+                        c.GetSettings().Set("Transport.AzureStorageQueue.DisablePublishSubscribe", true);
```
SeanFeldman, line 41: Should we use the transport's configuration API `EnableMessageDrivenPubSubCompatibilityMode()` instead of reaching out in the setting?

src/Transport/AzureStorageQueueInfrastructure.cs
```diff
@@ -88,6 +105,37 @@ internal AzureStorageQueueInfrastructure(SettingsHolder settings, string connect
                 };
             }
 
+            var subscriptionTableName = GetSubscriptionTableName(settings);
+
+            object publishSubscribe = new { NativePublishSubscribeIsEnabled = false, };
+            subscriptionStore = new NoOpSubscriptionStore();
+            if (PublishSubscribeIsEnabled())
+            {
+                object subscriptionPersistenceCachingSection = new { IsEnabled = false };
```
SeanFeldman, line 114: nit: `object`? I thought `var` would suffice.

src/Transport/AzureStorageQueueInfrastructure.cs
```diff
@@ -260,9 +301,19 @@ public override string ToTransportAddress(LogicalAddress logicalAddress)
...
+                await EnsureSubscriptionTableExists(cloudTableClientProvider.Client, GetSubscriptionTableName(settings)).ConfigureAwait(false);
+            }
+            await nativeDelayedDelivery.Start().ConfigureAwait(false);
+        }
+
+        static async Task EnsureSubscriptionTableExists(CloudTableClient cloudTableClient, string subscriptionTableName)
         {
-            return nativeDelayedDelivery.Start();
+            var subscriptionTable = cloudTableClient.GetTableReference(subscriptionTableName);
+            await subscriptionTable.CreateIfNotExistsAsync().ConfigureAwait(false);
```
SeanFeldman, line 316: A single async operatoin that could be returned as a task, right?

src/Transport/Config/Extensions/AccountRoutingSettings.cs
```diff
@@ -22,12 +24,21 @@ public AccountInfo AddAccount(string alias, string connectionString)
             return accounts.Add(alias, connectionString);
         }
 
+        // TODO: adjust as needed
```
SeanFeldman, line 27: TODO is no longer needed, right?
```suggestion
```

SeanFeldman (approved):
Approving with one suggestion and a few comments.

src/AcceptanceTests/PubSub/When_migrating_publisher_first.cs
danielmarbach: No that wouldn't work. we are disabling the native pubsub mode. Compatibility mode is something different

src/Transport/AzureStorageQueueInfrastructure.cs
danielmarbach: It could yes although here it doesn't really matter at all since it is not perf sensitive. Plus we have it this way on master too

src/Transport/AzureStorageQueueInfrastructure.cs
```diff
@@ -88,6 +105,37 @@ internal AzureStorageQueueInfrastructure(SettingsHolder settings, string connect
                 };
             }
 
+            var subscriptionTableName = GetSubscriptionTableName(settings);
+
+            object publishSubscribe = new { NativePublishSubscribeIsEnabled = false, };
```
danielmarbach, line 110:
```suggestion
            var publishSubscribe = new { NativePublishSubscribeIsEnabled = false, };
```

src/Transport/AzureStorageQueueInfrastructure.cs
danielmarbach, line 114:
```suggestion
                var subscriptionPersistenceCachingSection = new { IsEnabled = false };
```

danielmarbach: it has to be object so that we can reassign different anonymous types to it

PR #563, opened 2021-03-15
GitHubSync update - release-10.0

particularbot6:
This is an automated synchronization PR.

The following source template repositories were used:
* particular/RepoStandards

PR #564, opened 2021-03-16
Revert "GitHubSync update - release-10.0"

WilliamBZA:
Reverts Particular/NServiceBus.AzureStorageQueues#563


danielmarbach:
@WilliamBZA Thanks. From our perspective we need the beta tag still. So, technically if you require the auto-merge config those could stay.

WilliamBZA:
Ok, added the dependabot files back

danielmarbach:
In the meantime we have smoke tested the last changes and found out we don't need the beta tag anymore. Sorry for messing with you. we can close this now

PR #570, opened 2021-03-18
Connection string option in new api

tmasternak:


mauroservienti:
Replaced by https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/577

PR #614, opened 2021-05-10
Aligning Task.Run and OperationCanceledException handling

DavidBoike:


DavidBoike:
Recreating PR so TeamCity will see it.

PR #615, opened 2021-05-10
Aligning Task.Run and OperationCanceledException handling

DavidBoike:
Replaces #614

### DIFF
--- a/src/Transport/AtLeastOnceReceiveStrategy.cs
+++ b/src/Transport/AtLeastOnceReceiveStrategy.cs
@@ -30,9 +30,18 @@ namespace NServiceBus.Transport.AzureStorageQueues
 
                 await retrieved.Ack(cancellationToken).ConfigureAwait(false);
             }
-            catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+            catch (OperationCanceledException oce)
             {
                 // Graceful shutdown
+                if (cancellationToken.IsCancellationRequested)
+                {
+                    Logger.Debug("Message processing cancelled. Rolling back transaction.", oce);
+                }
+                else
+                {
+                    Logger.Warn("OperationCanceledException thrown. Rolling back transaction.", oce);
+                }
+
             }
             catch (LeaseTimeoutException)
             {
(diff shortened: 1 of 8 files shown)


src/Transport/AtLeastOnceReceiveStrategy.cs
```diff
@@ -30,9 +30,18 @@ public override async Task Receive(MessageRetrieved retrieved, MessageWrapper me
 
                 await retrieved.Ack(cancellationToken).ConfigureAwait(false);
             }
-            catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+            catch (OperationCanceledException oce)
```
andreasohlund, line 33: Nitpick: We use `ex` in the other places

andreasohlund (approved):

src/Transport/AtLeastOnceReceiveStrategy.cs
DavidBoike: In this file there's an OCE captured within an existing catch(Exception ex) block so "ex" is not available, I went for keeping all the OCEs consistent within the file.

src/Transport/AtLeastOnceReceiveStrategy.cs
```diff
@@ -30,9 +30,18 @@ public override async Task Receive(MessageRetrieved retrieved, MessageWrapper me
...
             {
                 // Graceful shutdown
+                if (cancellationToken.IsCancellationRequested)
+                {
+                    Logger.Debug("Message processing cancelled. Rolling back transaction.", oce);
```
adamralph, line 38: Where is the rollback happening?

DavidBoike: Yeah, it's not explicitly. Just letting the lease expire, as a Nack would require another network call when we're already cancelling. I'll remove the "Rolling back transaction" part of the message in a new PR.

DavidBoike: @adamralph See https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/616

PR #628, opened 2021-05-22
fix OCE catching, Task.Run(), etc.

adamralph:


DavidBoike (approved):

src/Transport/MessageReceiver.cs
```diff
@@ -81,20 +83,22 @@ public async Task StopReceive(CancellationToken cancellationToken = default)
 
             using (cancellationToken.Register(() => messageProcessingCancellationTokenSource?.Cancel()))
             {
+                await Task.WhenAll(messagePumpTasks).ConfigureAwait(false);
+
```
adamralph, line 87: I think we need to do this here to ensure no message pump task is still running when we check the concurrency limiter, because otherwise a message pump task could start another message processing task after we check the concurrency limiter.

I think we also need it to ensure that no message pump task tries to use the message pump cancellation token after the token source is disposed at the end of this method, although that's probably quite a wide race condition.

README.md
```diff
@@ -7,8 +7,6 @@ The Azure Storage Queues transport for NServiceBus enables the use of the Azure
 * [Azure Transport](https://docs.particular.net/nservicebus/azure-storage-queues/)
 * [Samples](https://docs.particular.net/samples/azure/storage-queues/)
 
-## How to Test Locally
+## How to test locally
 
-Follow these steps to run the acceptance tests locally:
-* Add a new environment variable `Transport.UseSpecific` with the value `AzureStorageQueueTransport`
-* Add a new environment variable `AzureStorageQueueTransport.ConnectionString` containing a connection string to your Azure storage account or use `UseDevelopmentStorage=true` to use the [Azure Storage Emulator](https://azure.microsoft.com/en-us/documentation/articles/storage-use-emulator/) (make sure to start it before you run the tests).
+To run the tests locally, add a new environment variable `AzureStorageQueueTransport_ConnectionString` containing a connection string to your Azure storage account or the connection string `UseDevelopmentStorage=true` to use the [Azurite emulator](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azurite) (ensure it is started before you run the tests).
```
adamralph, line 12: This isn't entirely sufficient, since a few tests still fail with:

> Couldn't find an environment variable 'AzureStorageQueueTransport_ConnectionString_2' with Azure Storage connection string.

But I'm not sure what to do with that. The changes I've made here are at least a step in right direction.

andreasohlund (approved):

PR #698, opened 2021-09-22
The delay calculation moved to dispatcher

SzymonPobiega:
To simplify the delayed delivery interface

### DIFF
--- a/src/Transport/Dispatcher.cs
+++ b/src/Transport/Dispatcher.cs
@@ -8,22 +8,25 @@
     using System.Threading;
     using System.Threading.Tasks;
     using Azure.Transports.WindowsAzureStorageQueues;
+    using DelayedDelivery;
+    using DeliveryConstraints;
     using Extensibility;
     using global::Azure.Storage.Queues;
     using Logging;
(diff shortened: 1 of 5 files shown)


src/Transport/Dispatcher.cs
```diff
@@ -74,9 +77,15 @@ public async Task Send(UnicastTransportOperation operation, CancellationToken ca
                 Logger.DebugFormat("Sending message (ID: '{0}') to {1}", operation.Message.MessageId, operation.Destination);
             }
 
-            var dispatchDecision = await shouldSend(operation, cancellationToken).ConfigureAwait(false);
-            if (dispatchDecision == false)
+            var delay = GetVisibilityDelay(operation.DeliveryConstraints);
+            if (delay != null)
             {
+                if (FirstOrDefault<DiscardIfNotReceivedBefore>(operation.DeliveryConstraints) != null)
```
mikeminutillo, line 83: This check runs even if delayed delivery is disabled. This could lead to confusion when the user gets this error message, fixes it, and is then told that delayed delivery won't work anyway.

mikeminutillo (approved):

PR #742, opened 2021-11-25
Remove usage of obsoleted address translation method

timbussmann:
The current base PR still uses ToTransportAddress from the TransportDefinition to create the receivers. When the next TF will go and remove the implementation in TransportDefiniton this will cause problems because the receivers are built before the infrastructure.

This PR moves the building of the receivers inside the Infrastructure so that the usage of the address translation is only used once the infrastructure type has been created.

Not sure it's worthwhile doing this change, because it looks like ToTransportAddress still requires some stuff from TransportDefinition internally anyway, which is heavily used within the Initialze method, so it won't be that easy to remove other ongoing address translation from TransportDefinition anyway.

### DIFF
--- a/src/Transport/AzureStorageQueueInfrastructure.cs
+++ b/src/Transport/AzureStorageQueueInfrastructure.cs
@@ -1,16 +1,15 @@
 ﻿namespace NServiceBus.Transport.AzureStorageQueues
 {
-    using System.Collections.ObjectModel;
+    using System.Collections.Generic;
     using System.Threading;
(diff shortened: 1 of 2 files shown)


timbussmann:
discussed with Andreas and we agreed while this change seems good, it still won't allow us to fully remove address translation from `TransportDefinition` anyway. So the easier way might be to just make `ToTransportAddress` private in the next major to minimize the refactoring risk for this dependency update.

PR #754, opened 2022-01-18
Obsolete APIs for next major

aleksandr-samila:
Restored obsolete APIs `DisableTimeoutManager`  which are being removed

### DIFF
--- a/src/Transport/obsoletes-v11.cs
+++ b/src/Transport/obsoletes-v11.cs
@@ -1,5 +1,6 @@
 ﻿namespace NServiceBus
 {
+    using System;
     using global::Azure.Storage.Queues;
     using Microsoft.Azure.Cosmos.Table;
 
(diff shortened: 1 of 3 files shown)


src/Transport/obsoletes-v11.cs
```diff
@@ -19,4 +20,37 @@ partial class AccountRoutingSettings
...
+
+    /// <summary>Configures native delayed delivery.</summary>
+    public partial class DelayedDeliverySettings
+    {
+        /// <summary>
+        /// Disables the Timeout Manager for the endpoint. Before disabling ensure there all timeouts in the timeout store
+        /// have been processed or migrated.
+        /// </summary>
+        [ObsoleteEx(
+            TreatAsErrorFromVersion = "11.0",
```
HEskandari, line 32: hmmm shouldn't this be version 10? The current release tracking NSBv7 is version 9. The next version tracking NSBv8 would be 10. Since this is throwing an exception, it should error in v10, and be removed in v11, right? Same for below.

HEskandari: The source of confusion was the Release page on GH which showed version 9 to be the latest. This is correct.

HEskandari (approved):

mikeminutillo (approved):

PR #895, opened 2022-10-17
Replaced Microsoft.Azure.Cosmos.Table references with Azure.Data.Table

abparticular:


src/AcceptanceTests/NServiceBus.Transport.AzureStorageQueues.AcceptanceTests.csproj
```diff
@@ -4,22 +4,24 @@
...
     <PackageReference Include="GitHubActionsTestLogger" Version="2.0.1" />
     <PackageReference Include="Azure.Storage.Queues" Version="12.11.1" />
     <PackageReference Include="Azure.Storage.Blobs" Version="12.14.0" />
-    <PackageReference Include="Microsoft.Azure.Cosmos.Table" Version="1.0.8" />
     <PackageReference Include="NServiceBus.AcceptanceTests.Sources" Version="8.0.0-rc.3" />
     <PackageReference Include="NServiceBus.Newtonsoft.Json" Version="3.0.0-rc.3" />
     <PackageReference Include="NUnit3TestAdapter" Version="4.2.1" />
     <PackageReference Include="NUnit" Version="3.13.3" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.3.2" />
+    <PackageReference Include="System.Configuration.ConfigurationManager" Version="6.0.1" />
```
danielmarbach, line 24: Out of curiosity. Why is this needed?

src/AcceptanceTests/Sending/When_delaying_messages_natively.cs
```diff
@@ -1,70 +1,69 @@
...
-        CloudTable delayedMessagesTable;
+        TableClient delayedMessagesTable;
 
         [SetUp]
         public async Task SetUpLocal()
         {
-            delayedMessagesTable = CloudStorageAccount.Parse(Utilities.GetEnvConfiguredConnectionString()).CreateCloudTableClient().GetTableReference(SenderDelayedMessagesTable);
-            if (await delayedMessagesTable.ExistsAsync().ConfigureAwait(false))
+            var tableServiceClient = new TableServiceClient(Utilities.GetEnvConfiguredConnectionString());
+            delayedMessagesTable = tableServiceClient.GetTableClient(SenderDelayedMessagesTable);
```
danielmarbach, line 27: How about `delayedMessagesTableClient`?

src/AcceptanceTests/Sending/When_delaying_messages_natively.cs
```diff
@@ -1,70 +1,69 @@
...
+
+            var table = await tableServiceClient.QueryAsync(t => t.Name.Equals(SenderDelayedMessagesTable), 1).FirstOrDefaultAsync().ConfigureAwait(false);
```
danielmarbach, line 29: Should we make a comment here that [there is no exists](https://github.com/Azure/azure-sdk-for-net/issues/28392)?

danielmarbach: Now that I look at the code wouldn't it be more "performant" to delete and recreate the table?

src/AcceptanceTests/Sending/When_delaying_messages_natively.cs
```diff
@@ -195,6 +190,25 @@ public class MyMessage : IMessage
             public Guid Id { get; set; }
         }
 
+        class CaptureSendingRequestsPolicy : HttpPipelinePolicy
```
danielmarbach, line 193: Would it help here to quickly describe why we need this?

src/AcceptanceTests/Sending/When_delaying_messages_natively_with_send_only.cs
```diff
@@ -17,12 +17,16 @@ public class When_delaying_messages_natively_with_send_only : NServiceBusAccepta
...
+            var table = await tableServiceClient.QueryAsync(t => t.Name.Equals(SenderDelayedMessagesTable), 1).FirstOrDefaultAsync().ConfigureAwait(false);
+            if (table != null)
             {
-                foreach (var dte in await delayedMessagesTable.ExecuteQuerySegmentedAsync(new TableQuery(), null).ConfigureAwait(false))
+                await foreach (var dte in delayedMessagesTable.QueryAsync<TableEntity>())
                 {
-                    await delayedMessagesTable.ExecuteAsync(TableOperation.Delete(dte)).ConfigureAwait(false);
+                    var result = await delayedMessagesTable.DeleteEntityAsync(dte.PartitionKey, dte.RowKey).ConfigureAwait(false);
+                    Assert.That(result.IsError, Is.False, "Error {0}:{1} trying to delete {2}:{3} from {4} table", result.Status, result.ReasonPhrase, dte.PartitionKey, dte.RowKey, SenderDelayedMessagesTable);
                 }
```
danielmarbach, line 30: Same comment as above

src/AcceptanceTests/Sending/When_dispatching_fails.cs
```diff
@@ -37,63 +33,40 @@ public void Should_log_queue_related_error_if_queue_doesnt_exist()
...
+                EndpointSetup(new CustomizedServer(transport), (cfg, rd) => { });
             }
         }
 
         public class MyMessage : IMessage
         {
         }
 
-        public class MyMessageHandler : IHandleMessages<MyMessage>
+        class FailSendingRequestsPolicy : HttpPipelinePolicy
```
danielmarbach, line 65: Would it help to describe why we are doing this here as a comment? Maybe something along the lines of

> We are intercepting actually HTTP calls to simulate SDK failures

src/AcceptanceTests/Utilities.cs
```diff
@@ -54,5 +85,7 @@ public static AzureStorageQueueTransport SetTransportDefaultTestsConfiguration(A
 
             return transport;
         }
+
+        public static bool Contains(this string source, string subString, StringComparison comparison) => source.IndexOf(subString, comparison) >= 0;
```
danielmarbach, line 89: If this is being used somewhere, would it be better placed under something like `StringExtensions`? And why not directly using [`String.Contains`](https://learn.microsoft.com/en-us/dotnet/api/system.string.contains)?

src/Tests/ApprovalFiles/APIApprovals.ApproveAzureStorageQueueTransport.approved.txt
```diff
@@ -18,13 +18,13 @@ namespace NServiceBus
             " overload that accepts a QueueServiceClient instance. Will be treated as an erro" +
             "r from version 12.0.0. Will be removed in version 13.0.0.", false)]
```
danielmarbach, line 19: We probably also need to remove some of those obsoletes and double-check the version numbers. We can do that in a different PR, though.

danielmarbach (commented):
Initial set of comments. Will continue later but at least we have some should I not have further time today

src/Transport/DelayDelivery/NativeDelayDeliveryPersistence.cs
```diff
@@ -78,10 +68,10 @@ public Task ScheduleAt(UnicastTransportOperation operation, DateTimeOffset date,
             };
 
             delayedMessageEntity.SetOperation(operation);
-            return delayedMessageStorageTable.ExecuteAsync(TableOperation.Insert(delayedMessageEntity), null, null, cancellationToken);
+            return delayedMessageStorageTableClient.UpsertEntityAsync(delayedMessageEntity, cancellationToken: cancellationToken);
```
danielmarbach, line 71: Is Upsert the right approach here? Insert assumes the entity does not exist and would throw if it did. Upsert would replace it. Should we use `AddEntityAsync instead? 

https://learn.microsoft.com/en-us/dotnet/api/azure.data.tables.tableclient.addentityasync

Maybe changing the behavior here is the right thing to do but we need to think this through.

src/Transport/NServiceBus.Transport.AzureStorageQueues.csproj
```diff
@@ -13,14 +13,15 @@
...
+    <PackageReference Include="Azure.Data.Tables" Version="[12.6.1, 13.0.0)" />
     <PackageReference Include="Azure.Storage.Queues" Version="[12.11.1, 13.0.0)" />
     <PackageReference Include="Azure.Storage.Blobs" Version="[12.13.1, 13.0.0)" />
-    <PackageReference Include="Microsoft.Azure.Cosmos.Table" Version="[1.0.8, 2.0.0]" />
     <PackageReference Include="NServiceBus" Version="[8.0.0-rc.3, 9)" />
     <PackageReference Include="Fody" Version="6.6.4" PrivateAssets="All" />
     <PackageReference Include="Janitor.Fody" Version="1.9.0" PrivateAssets="All" />
     <PackageReference Include="Obsolete.Fody" Version="5.3.0" PrivateAssets="All" />
     <PackageReference Include="Particular.Packaging" Version="2.2.0" PrivateAssets="All" />
+    <PackageReference Include="System.Linq.Async" Version="6.0.1" />
```
danielmarbach, line 24: Wouldn't this package only be needed for .NET Framework, but can be removed for .NET 6? Where do we need these extensions?

src/Transport/PubSub/SubscriptionStore.cs
```diff
@@ -26,59 +25,53 @@ public async Task<IEnumerable<string>> GetSubscribers(Type eventType, Cancellati
...
+            var entity = new SubscriptionEntity
             {
                 Topic = TopicName.From(eventType),
                 Endpoint = endpointName,
                 Address = address.ToString()
-            });
-            return table.ExecuteAsync(operation, cancellationToken);
+            };
+
+            return tableClient.UpsertEntityAsync(entity, TableUpdateMode.Merge, cancellationToken: cancellationToken);
```
danielmarbach, line 64: TableUpdateMode.Replace?

src/Transport/Utils/StorageUri.cs
```diff
@@ -0,0 +1,299 @@
...
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+// </copyright>
+//-----------------------------------------------------------------------
+
+namespace NServiceBus.Transport.AzureStorageQueues.Utils
```
danielmarbach, line 18: Should we also leave the original namespace?

danielmarbach: Or if we tweak things we could also remove the #if that never applies to us

src/Transport/Utils/StorageUri.cs
```diff
@@ -0,0 +1,299 @@
...
+    using System.Collections.Generic;
+    using System.Globalization;
+    using System.Linq;
+
+#if WINDOWS_RT
+    using Windows.Foundation.Metadata;
+#endif
+
+    /// <summary>
+    /// Contains the URIs for both the primary and secondary locations of a Microsoft Azure Storage resource.
```
danielmarbach, line 30: Maybe add a github link to the original file in the comment?

src/Transport/DelayDelivery/DelayedMessageEntity.cs
```diff
@@ -2,28 +2,23 @@
...
         public string Destination { get; set; }
         public byte[] Body { get; set; }
         public string MessageId { get; set; }
         public string Headers { get; set; }
 
-        static string Serialize<T>(T value)
-        {
-            return SimpleJson.SimpleJson.SerializeObject(value);
-        }
+        static string Serialize<T>(T value) => SimpleJson.SimpleJson.SerializeObject(value);
```
danielmarbach, line 19: Azure.Data.Table also brings in System.Text.Json? Should we verify if we can dump simplejson already or is that too much of a risk? (Can also be done as a dedicated work item/PR)?

danielmarbach (commented):
Last round of reviews

src/AcceptanceTests/Sending/When_delaying_messages_natively.cs
abparticular: I don't know. I did think about that when updating the code, but there only appears to be a couple of tests that use the table, and they only add a single row by my reading.  Maybe the overhead of dropping/creating tables is higher?  Either way, due to the low usage in an acceptance test I decided against doing a performance analysis and just kept the original logic.

src/AcceptanceTests/Utilities.cs
abparticular: The standard `String.Contains` doesn't allow for case-insensitive comparisons.  It is used [here](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/ff943c4b03e5c96328b74046bc835060ce9cd802/src/AcceptanceTests/Sending/When_delaying_messages_natively.cs#L53-L54).

src/Transport/AzureStorageQueueTransport.cs
```diff
@@ -492,7 +487,7 @@ public TimeSpan MaximumWaitTimeWhenIdle
             {
                 Guard.AgainstNull(nameof(QueueNameSanitizer), value);
 
-                Func<string, string> queueNameSanitizerWrapper = entityName =>
+                string queueNameSanitizerWrapper(string entityName)
```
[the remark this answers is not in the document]
abparticular: No, it references `value`

src/Transport/NServiceBus.Transport.AzureStorageQueues.csproj
abparticular: No it's needed for .NET 6 too, for LINQ methods that take an IAsyncEnumerable, e.g. [here](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/ff943c4b03e5c96328b74046bc835060ce9cd802/src/Transport/PubSub/SubscriptionStore.cs#L49-L50)

src/Transport/DelayDelivery/NativeDelayDeliveryPersistence.cs
abparticular: The behaviour change wasn't deliberate here, I'll use the Add method instead.

src/Transport/Utils/StorageUri.cs
abparticular: I wasn't sure about what was needed to get the licensing correct to be honest, it was one of the things I wanted to talk about.  In terms of the namespace, I changed it because I had to change the actual code a bit to get it working for us - it's not a straight cut and paste of the original code.

src/Transport/DelayDelivery/DelayedMessageEntity.cs
abparticular: I'm happy to look into that, but would prefer to do it separately to this issue.

src/AcceptanceTests/NServiceBus.Transport.AzureStorageQueues.AcceptanceTests.csproj
abparticular: Use of `ConfigurationErrorsException` [here](https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/ff943c4b03e5c96328b74046bc835060ce9cd802/src/AcceptanceTests/Utilities.cs#L71)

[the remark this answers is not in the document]
abparticular: GitHub is weird to me... this seems to be a comment on the same code as the previous question, although it doesn't show up on the Files Changed view?

src/Transport/DelayDelivery/DelayedMessagesPoller.cs
```diff
@@ -121,13 +122,11 @@ async Task SpinOnce(CancellationToken cancellationToken)
 
             Logger.Debug($"Polling for delayed messages at {now}.");
 
-            var query = new TableQuery<DelayedMessageEntity>
-            {
-                FilterString = $"(PartitionKey le '{DelayedMessageEntity.GetPartitionKey(now)}') and (RowKey le '{DelayedMessageEntity.GetRawRowKeyPrefix(now)}')",
-                TakeCount = DelayedMessagesProcessedAtOnce // max batch size
```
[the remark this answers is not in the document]
abparticular: I thought the paging would be handled automagically by QueryAsync but on closer inspection I think you're correct.

src/Transport/Config/QueueAddress.cs
```diff
@@ -20,39 +20,33 @@ public QueueAddress(string queueName, string alias)
...
+        public bool Equals(QueueAddress other) =>
+            string.Equals(QueueName, other.QueueName, StringComparison.OrdinalIgnoreCase) &&
+            string.Equals(Alias, other.Alias, StringComparison.OrdinalIgnoreCase);
 
-        static bool IsQueueNameValid(string queueName)
-        {
-            return string.IsNullOrWhiteSpace(queueName) == false;
-        }
+        public static QueueAddress Parse(string inputQueue, bool allowConnectionStringForBackwardCompatibility = false) =>
+            TryParseInternal(inputQueue, allowConnectionStringForBackwardCompatibility, out var q, error => throw new FormatException(error))
```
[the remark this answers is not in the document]
timbussmann: no I don't remember written guidelines about this 🤔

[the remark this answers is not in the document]
danielmarbach: The closest I could find with the help of Simon was https://discuss.particular.net/t/do-not-throw-system-exception/414

danielmarbach:
Rebased with force push

danielmarbach (approved):

danielmarbach (approved):

danielmarbach:
I think this is ready to go.

src/Transport/Config/QueueAddress.cs
```diff
@@ -20,39 +20,35 @@ public QueueAddress(string queueName, string alias)
...
-        public static QueueAddress Parse(string value, bool allowConnectionStringForBackwardCompatibility = false)
-        {
-            if (TryParse(value, allowConnectionStringForBackwardCompatibility, out var q) == false)
-            {
-                throw new ArgumentException("Value cannot be parsed", nameof(value));
-            }
+        public static QueueAddress Parse(string inputQueue, bool allowConnectionStringForBackwardCompatibility = false) =>
+            TryParseInternal(inputQueue, allowConnectionStringForBackwardCompatibility, out var q, static error => throw new FormatException(error))
+                ? q.Value
+                : throw new FormatException($"Cannot parse '{inputQueue}' as an address");
```
lailabougria, line 30: Generally speaking, I'd say there's a risk when changing the type of exception we throw. But I guess since this scenario would affect the startup of the endpoint we're mostly good. I do wonder though what the thought process was for changing it from Argument to FormatException?

danielmarbach: See the discussion https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/895#discussion_r1001077312
This ultimately lead to the exception guidelines we have been writing. Following the original assumption is that that calling code is most likely to have a general try catch I guess change would be OKish. That being said if we want to stay compatible even between major versions we could roll this back.

Otherwise having it mentioned in the upgrade guide might be good enough given the low risk of this exception being an actual problem.

src/Transport/DelayDelivery/DelayedMessagesPoller.cs
```diff
@@ -195,6 +194,18 @@ async Task DeleteAndDispatch(DelayedMessageEntity delayedMessage, CancellationTo
             }
         }
 
+        /// <returns>true if delete succeeded, false otherwise</returns>
+        async Task<bool> TryDeleteDelayedMessage(DelayedMessageEntity delayedMessage, CancellationToken cancellationToken)
+        {
+            var response = await delayedDeliveryTableClient.DeleteEntityAsync(delayedMessage.PartitionKey, delayedMessage.RowKey, cancellationToken: cancellationToken).ConfigureAwait(false);
+            if (response.IsError)
+            {
+                Logger.Warn($"Failed to delete the delayed message with PartitionKey:'{delayedMessage.PartitionKey}' RowKey: '{delayedMessage.RowKey}' message ID: '{delayedMessage.MessageId}'");
```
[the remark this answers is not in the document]
danielmarbach: As far as I remember @abparticular has tested the behavior and validated the assumption (at least that is what I seem to remember from the conversation of one of the sync calls, but my memories are not the best)

src/Transport/Config/ICloudTableClientProvider.cs
```diff
@@ -1,16 +0,0 @@
-﻿namespace NServiceBus.Transport.AzureStorageQueues
```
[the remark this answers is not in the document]
lailabougria: Ah never mind :)

lailabougria (approved):

src/Transport/DelayDelivery/DelayedMessagesPoller.cs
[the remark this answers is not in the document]
abparticular: My original change to this code wasn't passing in the ETag of the entity to delete.  @danielmarbach correctly picked this up as a change in behaviour as the previous Cosmos Table API would automatically include the ETag in a delete operation (which is what I confirmed).  I modified the new implementation to also pass in the ETag.

So the deletes can possibly fail due to concurrency issues.  I don't know how likely that is to occur though.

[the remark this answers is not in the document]
danielmarbach: @abparticular and I have discussed this again and came to the conclusion that since we are passing in the ETag and the operation will return 412 under the covers that will be turned into a `RequestFailedException` thrown to the caller code. For all other errors that might not be throwing it is good to have the `IsError` check with a log warn in there

PR #908, opened 2022-11-11
Client update and small cleanups

danielmarbach:


src/Transport/PubSub/SubscriptionStore.cs
```diff
@@ -104,5 +104,7 @@ static IEnumerable<Type> GenerateMessageHierarchy(Type messageType)
         }
 
         readonly ConcurrentDictionary<Type, string[]> eventTypeToTopicListMap = new();
+
+        static readonly IEnumerable<string> SelectedSubscriptionColumns = new[] { nameof(SubscriptionEntity.Address) };
```
abparticular, line 108:
```suggestion
        static readonly IEnumerable<string> AddressColumnFilter = new[] { nameof(SubscriptionEntity.Address) };
```
(Obviously would also need to make a corresponding change at usage site)

abparticular (approved):

src/Transport/PubSub/SubscriptionStore.cs
danielmarbach: Done. Good suggestion

lailabougria (approved):

Issue #991, opened 2023-03-25
Improve handling of messages that are close to the max body limit


danielmarbach:
#### Is your improvement related to a problem? Please describe.

This is a tricky situation and we might not be able to improve the situation fully. Today the story when we receive a message that is close to the 64 KB size limit is problematic. When we receive such a large message and either send it to the audit queue or want to send it to the error queue, we enrich the message with more headers. Once that happens we reach the body size limit and then that messages is indefinitely retried. Because there is no delay and it also cannot be handled by the delayed delivery infrastructure, such a message ends up churning away significant resources of the endpoint which might even cause the endpoint to completely stall. 

We already have solutions like the databus but the problem is once you got such a message things problematic and you have to somehow manually receive messages from the queue until you reach the poison message to delete it with the corresponding pop receipt while requeue-ing all other messages. 

#### Describe the suggested solution

- Some sort of oversized message handler like we had once in ASB?
- Native databus support inside the transport similar to what SQS does with S3
- Log warn/error and at least use the client side native delivery to move the message with backoff to give the users awareness of this problem
- Eventually backup the message somehow and ack it?
- Allow specifying a DLQ for the endpoint and move the message there (my favorite)

See also https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/issues/138#issuecomment-781513452

PR #1002, opened 2023-05-12
Make sure messages close to the size limit can be moved as is to the error queue

danielmarbach:
Fixes https://github.com/Particular/NServiceBus.AzureStorageQueues/issues/991

I rebased my previous work that I did for a customer support request and figured I send in the PR


WilliamBZA (approved):

udidahan:
Might it be helpful if the transport exposed the max size of message it can handle, and then Core could do something more generic/intelligent with that?

udidahan:
OK. Thanks.

PR #1004, opened 2023-05-18
Add the ability to try to at least add minimal headers for messages being sent to the error queue

boblangley:
A PR against #1002 

This PR adds the ability to at least try to add the `FailedQ` and `ExceptionType` headers before resorting to sending the raw message to the error queue.

The FailedQ header is required by ServiceControl or it will end up being a failed error import.
The ExceptionType is highly desireable as it will allow grouping. The failure group view by exception type is the default failed message view in ServicePulse, which is the common way for users to interact with failed messages. A message without this header would only be visible in the all failed messages view which would make failed messages difficult to discover for customers with many active failed messages.

### DIFF
--- a/src/AcceptanceTests/Receiving/When_receiving_large_message.cs
+++ b/src/AcceptanceTests/Receiving/When_receiving_large_message.cs
@@ -11,12 +11,14 @@ namespace NServiceBus.Transport.AzureStorageQueues.AcceptanceTests
     using global::Newtonsoft.Json;
     using NServiceBus.AcceptanceTests;
     using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NServiceBus.Faults;
     using NUnit.Framework;
 
     public class When_receiving_large_message : NServiceBusAcceptanceTest
     {
         [Test]
-        public async Task Should_consume_it()
+        // msg size very close  to limit , so when an exception happens, we  cannot add error queue headers, we send only the raw message body to the error queue
+        public async Task Should_consume_it_without_the_error_headers_when_message_size_very_close_to_limit()
         {
             var ctx = await Scenario.Define<Context>()
                 .WithEndpoint<Receiver>(b =>
@@ -56,11 +58,62 @@ namespace NServiceBus.Transport.AzureStorageQueues.AcceptanceTests
                 .WithEndpoint<ErrorSpy>()
                 .Done(c => c.MessageMovedToTheErrorQueue)
                 .Run();
+            Assert.IsFalse(ctx.IsFailedQHeaderPresent);
+            Assert.IsFalse(ctx.IsExceptionTypeHeaderPresent);
+        }
+
+        [Test]
+        // msg size close  to limit , so when an exception happens, we  can only add 2 headers before sending to error queue
+        public async Task Should_consume_it_with_only_two_error_headers_when_message_size_close_to_limit()
+        {
+            var ctx = await Scenario.Define<Context>()
+                .WithEndpoint<Receiver>(b =>
+                {
+                    b.When((bus, c) =>
+                    {
+                        var connectionString = Testing.Utilities.GetEnvConfiguredConnectionString();
+                        var queueClient = new QueueClient(connectionString, "receivinglargemessage-receiver");
+
+                        string contentCloseToLimits = new string('x', 33 * 1024);
+
+                        var message = new MyMessage { SomeProperty = contentCloseToLimits, };
+
+                        var messageSerialized = JsonConvert.SerializeObject(message, typeof(MyMessage), Formatting.Indented, new JsonSerializerSettings());
+
+                        string id = Guid.NewGuid().ToString();
+                        var wrapper = new MessageWrapper
+                        {
+                            Id = id,
+                            Body = Encoding.UTF8.GetBytes(messageSerialized),
+                            Headers = new Dictionary<string, string>
+                            {
+                         { Headers.EnclosedMessageTypes, $"{typeof(MyMessage).AssemblyQualifiedName}" },
+                         { Headers.MessageId, id },
+                         { Headers.CorrelationId, id },
+                         {TestIndependence.HeaderName, c.TestRunId.ToString()}
+                            }
+                        };
+
+                        var wrapperSerialized = JsonConvert.SerializeObject(wrapper, typeof(MessageWrapper), Formatting.Indented, new JsonSerializerSettings());
+
+                        var base64Encoded = Convert.ToBase64String(Encoding.UTF8.GetBytes(wrapperSerialized));
+
+                        return queueClient.SendMessageAsync(base64Encoded);
+                    }).DoNotFailOnErrorMessages();
+                })
+                .WithEndpoint<ErrorSpy>()
+                .Done(c => c.MessageMovedToTheErrorQueue)
+                .Run();
+            Assert.IsTrue(ctx.IsFailedQHeaderPresent);
+            Assert.IsTrue(ctx.IsExceptionTypeHeaderPresent);
         }
 
         class Context : ScenarioContext
         {
             public bool MessageMovedToTheErrorQueue { get; set; }
+            public bool IsFailedQHeaderPresent { get; set; }
+            public bool IsExceptionTypeHeaderPresent { get; set; }
+
         }
 
         class Receiver : EndpointConfigurationBuilder
@@ -99,7 +152,16 @@ namespace NServiceBus.Transport.AzureStorageQueues.AcceptanceTests
                     {
                         testContext.MessageMovedToTheErrorQueue = true;
                     }
-
+                    if (context.MessageHeaders.ContainsKey(FaultsHeaderKeys.FailedQ)
+                        && testRunId == testContext.TestRunId.ToString())
+                    {
+                        testContext.IsFailedQHeaderPresent = true;
+                    }
+                    if (context.MessageHeaders.ContainsKey("NServiceBus.ExceptionInfo.ExceptionType")
+                        && testRunId == testContext.TestRunId.ToString())
+                    {
+                        testContext.IsExceptionTypeHeaderPresent = true;
+                    }
                     return Task.CompletedTask;
                 }
 
--- a/src/AcceptanceTests/Receiving/When_receiving_large_message.cs
+++ b/src/AcceptanceTests/Receiving/When_receiving_large_message.cs
@@ -28,7 +28,8 @@ namespace NServiceBus.Transport.AzureStorageQueues.AcceptanceTests
                         var connectionString = Testing.Utilities.GetEnvConfiguredConnectionString();
                         var queueClient = new QueueClient(connectionString, "receivinglargemessage-receiver");
 
-                        string contentCloseToLimits = new string('x', 35 * 1024);
+                        //This value is fine tuned to ensure adding the 2 error headers make the message too large
+                        string contentCloseToLimits = new string('x', (35 * 1024) + 425);
 
                         var message = new MyMessage { SomeProperty = contentCloseToLimits, };
 
@@ -58,8 +59,9 @@ namespace NServiceBus.Transport.AzureStorageQueues.AcceptanceTests
                 .WithEndpoint<ErrorSpy>()
                 .Done(c => c.MessageMovedToTheErrorQueue)
                 .Run();
-            Assert.IsFalse(ctx.IsFailedQHeaderPresent);
-            Assert.IsFalse(ctx.IsExceptionTypeHeaderPresent);
+
+            Assert.IsFalse(ctx.IsFailedQHeaderPresent, "IsFailedQHeaderPresent");
+            Assert.IsFalse(ctx.IsExceptionTypeHeaderPresent, "IsExceptionTypeHeaderPresent");
         }
 
         [Test]
@@ -74,7 +76,7 @@ namespace NServiceBus.Transport.AzureStorageQueues.AcceptanceTests
                         var connectionString = Testing.Utilities.GetEnvConfiguredConnectionString();
                         var queueClient = new QueueClient(connectionString, "receivinglargemessage-receiver");
 
-                        string contentCloseToLimits = new string('x', 33 * 1024);
+                        string contentCloseToLimits = new string('x', (35 * 1024) + 400);
 
                         var message = new MyMessage { SomeProperty = contentCloseToLimits, };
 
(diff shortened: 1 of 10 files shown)


src/Transport/DefaultMessageEnvelopeUnwrapper.cs
```diff
@@ -36,6 +37,17 @@ public MessageWrapper Unwrap(QueueMessage rawMessage)
             return m;
         }
 
+        public BinaryData ReWrap(MessageWrapper wrapper)
```
boblangley, line 40: This needs to be moved out to somewhere else. We did this for convenience as we were initially working on a solution, but it doesn't make sense in the context of this class.

src/Transport/IMessageEnvelopeUnwrapper.cs
```diff
@@ -1,10 +1,12 @@
 namespace NServiceBus.Transport.AzureStorageQueues
 {
+    using System;
     using Azure.Transports.WindowsAzureStorageQueues;
     using global::Azure.Storage.Queues.Models;
 
     interface IMessageEnvelopeUnwrapper
     {
         MessageWrapper Unwrap(QueueMessage rawMessage);
+        BinaryData ReWrap(MessageWrapper wrapper);
```
boblangley, line 10: See above. Revert this.

boblangley:
This does raise a concern: In a situation where a customer has specified a custom unwrapper, the delayed or faulted message will be wrapped in the standard way by the transport and then may not become a poison message for the transport when a retry is attempted.

The solution is either to try the default unwrapper first, as implied but not fully implemented in [the snippet for writing a custom unwrapper](https://docs.particular.net/transports/azure-storage-queues/native-integration#custom-envelope-wrapper) OR if a custom unwrapper is defined always send just the raw message to the error queue, which would be difficult to retry.

src/AcceptanceTests/Receiving/When_receiving_large_message.cs
```diff
@@ -56,13 +58,62 @@ public async Task Should_consume_it()
...
+        {
+            var ctx = await Scenario.Define<Context>()
+                .WithEndpoint<Receiver>(b =>
+                {
+                    b.When((bus, c) =>
+                    {
+                        var connectionString = Testing.Utilities.GetEnvConfiguredConnectionString();
+                        var queueClient = new QueueClient(connectionString, "receivinglargemessage-receiver");
+
+                        string contentCloseToLimits = new string('x', 33 * 1024);
```
boblangley, line 77: We need to play with this value to tighten up the test.

boblangley:
Daniel, I requested your review on this WIP implementation in case you have some time to chime in. I tried to add some notes to hint about the deficiencies we already know about, and we haven't fully confirmed the implementation is fully correct yet, but I had hoped you would see the direction we are going and see if you could provide any feedback before we finish this up.

soujay (approved):

danielmarbach:
I was off a few days. I can try to look at it maybe tomorrow during travelling

src/AcceptanceTests/Receiving/When_receiving_large_message.cs
```diff
@@ -11,12 +11,14 @@ namespace NServiceBus.Transport.AzureStorageQueues.AcceptanceTests
...
     using NServiceBus.AcceptanceTests;
     using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NServiceBus.Faults;
     using NUnit.Framework;
 
     public class When_receiving_large_message : NServiceBusAcceptanceTest
     {
         [Test]
-        public async Task Should_consume_it()
+        // msg size very close  to limit , so when an exception happens, we  cannot add error queue headers, we send only the raw message body to the error queue
```
danielmarbach, line 20: What does the comment provide given the test name is already very descriptive?

src/AcceptanceTests/Receiving/When_receiving_large_message.cs
```diff
@@ -26,7 +28,8 @@ public async Task Should_consume_it()
                         var connectionString = Testing.Utilities.GetEnvConfiguredConnectionString();
                         var queueClient = new QueueClient(connectionString, "receivinglargemessage-receiver");
 
-                        string contentCloseToLimits = new string('x', 35 * 1024);
+                        //This value is fine tuned to ensure adding the 2 error headers make the message too large
```
danielmarbach, line 31: By looking at these precise numbers it looks like there is some sort of math involved we can do which got me wondering whether we can apply the same at runtime to determine whether we can take one code path or the other? My assumption is though that we tweaked these numbers for a very specific scenario in this test taking into account a well known serializer, header name and header value length etc, is that correct? What could influence these values to be suddenly off later, example test names?

src/AcceptanceTests/Receiving/When_receiving_large_message.cs
```diff
@@ -57,12 +60,63 @@ public async Task Should_consume_it()
...
                 .Run();
 
-            Assert.True(ctx.MessageMovedToTheErrorQueue);
+            Assert.IsFalse(ctx.IsFailedQHeaderPresent, "IsFailedQHeaderPresent");
+            Assert.IsFalse(ctx.IsExceptionTypeHeaderPresent, "IsExceptionTypeHeaderPresent");
+        }
+
+        [Test]
+        // msg size close  to limit , so when an exception happens, we  can only add 2 headers before sending to error queue
+        public async Task Should_consume_it_with_only_two_error_headers_when_message_size_close_to_limit()
```
danielmarbach, line 69: Same comment as above regarding the comment vs the test name

src/AcceptanceTests/Receiving/When_receiving_large_message.cs
```diff
@@ -57,12 +60,63 @@ public async Task Should_consume_it()
...
+        {
+            var ctx = await Scenario.Define<Context>()
+                .WithEndpoint<Receiver>(b =>
+                {
+                    b.When((bus, c) =>
+                    {
+                        var connectionString = Testing.Utilities.GetEnvConfiguredConnectionString();
+                        var queueClient = new QueueClient(connectionString, "receivinglargemessage-receiver");
+
+                        string contentCloseToLimits = new string('x', (35 * 1024) + 400);
```
danielmarbach, line 79: Same question as above

src/AcceptanceTests/Receiving/When_receiving_large_message.cs
```diff
@@ -57,12 +60,63 @@ public async Task Should_consume_it()
...
+
+                        var message = new MyMessage { SomeProperty = contentCloseToLimits, };
+
+                        var messageSerialized = JsonConvert.SerializeObject(message, typeof(MyMessage), Formatting.Indented, new JsonSerializerSettings());
```
danielmarbach, line 83: Could part of the logic here be extracted into a helper method and then be parametrized depending on the scenario? It seems an easy cleanup giving those two scenarios are in the same file

src/Transport/MessageRetrieved.cs
```diff
@@ -62,9 +65,38 @@ public async Task MoveToErrorQueue(CancellationToken cancellationToken = default
...
-            await inputQueue.DeleteMessageAsync(messageId, messagePopReceipt, cancellationToken).ConfigureAwait(false);
+            await TryMoveToErrorQueue(messageId, messagePopReceipt, rawMessage.Body, cancellationToken).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Moves the message without expiry to the error queue with minimal fault headers
+        /// </summary>
+        public async Task MoveToErrorQueue(ErrorContext context, CancellationToken cancellationToken = default)
+        {
+            var unwrappedMessage = await Unwrap(cancellationToken).ConfigureAwait(false);
```
danielmarbach, line 76: Given MessageRetrieved is a stateful thing would it make sense to assign the unwrapped message to a field so that unwrapping again would not be necessary? Because the order seems to be

1. Unwrap (if that fails move to error without headers) (poison handling)
2. If pipeline failed (can only happen if previous unwrap was successful) try move to to error either with all headers, partial headers or as is.

src/Transport/MessageRetrieved.cs
```diff
@@ -62,9 +65,38 @@ public async Task MoveToErrorQueue(CancellationToken cancellationToken = default
...
+
+            unwrappedMessage.Headers.Add(FaultsHeaderKeys.FailedQ, inputQueue.Name);
+            unwrappedMessage.Headers.Add("NServiceBus.ExceptionInfo.ExceptionType", context.Exception.GetType().FullName);
+
+            var body = ReWrap(unwrappedMessage);
+
+            await TryMoveToErrorQueue(rawMessage.MessageId, rawMessage.PopReceipt, body, cancellationToken).ConfigureAwait(false);
```
danielmarbach, line 83: Usually methods named `Try` follow a pattern of returning a bool indicating whether it was successful. This methods outcome is that it will be moved to the error queue. So why `Try` ?

src/Transport/MessageRetrieved.cs
danielmarbach: Maybe we can give the `MoveToError`  methods better names like `MoveToErrorAsIs` or `MoveToErrorWithoutModification` or `MoveToErrorWithoutErrorHeaders` or something like this to clearly distinguish the different cases`

src/Transport/MessageRetrieved.cs
```diff
@@ -99,11 +131,22 @@ public async Task Nack(CancellationToken cancellationToken = default)
             }
         }
 
+        BinaryData ReWrap(MessageWrapper wrapper)
```
danielmarbach, line 134: Uff this is allocation heavy... That being said I realize this is already a problem in the current implementation of the transport. So let's focus on the code reuse part here. It seems we do exactly the same here https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/80d292fd77adcca56981d574f8c5177be118333a/src/Transport/Dispatcher.cs#L129-L135. Is there a way we can align those paths?

src/Transport/MessageRetrieved.cs
```diff
@@ -62,9 +65,38 @@ public async Task MoveToErrorQueue(CancellationToken cancellationToken = default
...
+
+        /// <summary>
+        /// Moves the message without expiry to the error queue with minimal fault headers
+        /// </summary>
+        public async Task MoveToErrorQueue(ErrorContext context, CancellationToken cancellationToken = default)
+        {
+            var unwrappedMessage = await Unwrap(cancellationToken).ConfigureAwait(false);
+
+            unwrappedMessage.Headers.Add(FaultsHeaderKeys.FailedQ, inputQueue.Name);
+            unwrappedMessage.Headers.Add("NServiceBus.ExceptionInfo.ExceptionType", context.Exception.GetType().FullName);
```
danielmarbach, line 79: Maybe I'm overthinking things yet I want to mention it because it might triggers some good thoughts on your end. currently we have no way of knowing whether the repacking and everything is actually not going to be wasting a ton of resources. As an example when these scenarios are quite likely to happen anyway for messages that are super close to the limits we end up going through a multi staged attempt of trying to send it to the error queue which comes with computational costs, memory costs, storage operation costs and latency that might make it more likely to push across the visibility timeout.

So I wonder whether whether all this is really necessary (and I have no proof that it isn't either, I'm thinking out loud).

Already today we move poison messages to the error queue (and yes it is debatable whether that is a good idea but we do it and not just in ASQ). So the platform tooling needs to deal with such cases that do not have any headers. If we treat these oversize scenarios as rare and kind of binary that we can either send it with al the others or not and then have to send it raw we are addressing the poison pill scenarios quite nicely already without all the overhead costs mentioned above.

Another angle potentially worth considering is that I think we could estimate the required space of those two headers in base64 or non-base64 scenarios and then based on that choose one path or the other. Then we might be able to eliminate the overhead of try, fail, try again.

boblangley: > So the platform tooling needs to deal with such cases that do not have any headers.

I agree, but that is outside the scope of what this enhancement release can do, but should be raised so it can be prioritized. 

Your PR (which this is raised against) rightfully takes away the worst scenario that can cause the endpoint to essentially lock up.

However, the original solution, treats the message as "poison" which puts the customer into a position where none of our tooling really helps. It will show up as a failed error import with no chance that can be resolved without heavy manual intervention.

This at least provides a chance for normal retries.

> currently we have no way of knowing whether the repacking and everything is actually not going to be wasting a ton of resources.

If customers are regularly close to the 64kb limit and have consistent exceptions thrown in their handlers to the point that this becomes a serious resource drag perhaps we need to help them resolve that issue either by slimming down their message (data bus or attachments) or resolving the root instability.

src/Transport/MessageRetrieved.cs
boblangley: Usually yes, but you didn't use the word always :)

However, [we can rename it](https://martinfowler.com/bliki/TwoHardThings.html).

src/Transport/MessageRetrieved.cs
boblangley: Good idea. Will do.

src/Transport/Utils/MessageWrapperHelper.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace NServiceBus.Transport.AzureStorageQueues.Utils
+{
+    using System;
+    using System.IO;
+    using NServiceBus.Azure.Transports.WindowsAzureStorageQueues;
+
+    static class MessageWrapperHelper
```
boblangley, line 7: Since we pass in a serializer, perhaps this shouldn't be static, but instead a utility class that is instanced when the transport starts and passed to the dispatcher/receiver through the constructor?

src/Transport/MessageRetrieved.cs
```diff
@@ -55,16 +58,45 @@ public Task Ack(CancellationToken cancellationToken = default)
...
+
+            await MoveToErrorQueue(rawMessage.MessageId, rawMessage.PopReceipt, body, cancellationToken).ConfigureAwait(false);
+        }
+
+        public async Task MoveToErrorQueue(string messageId, string messagePopReceipt, BinaryData body, CancellationToken cancellationToken = default)
+        {
+            try
+            {
+                await errorQueue.SendMessageAsync(body, timeToLive: TimeSpan.FromSeconds(-1), cancellationToken: cancellationToken).ConfigureAwait(false);
+                // TODO: might not need this as the new SDK doesn't send a message by using the original message. Rather, copies the text only.
```
WilliamBZA, line 91: Can we investigate this todo so that it no longer lives in the code?

src/Transport/MessageRetrieved.cs
soujay: Applied the suggestions

src/Transport/Utils/MessageWrapperHelper.cs
soujay: I am comfortable in leaving it as it is. It this causes any issues in future; we can modify the code by removing the static type

src/Transport/MessageRetrieved.cs
boblangley: Since this existed prior to this PR (it was moved from a previous position) we should address the TODO separately.

PR #1023, opened 2023-06-05
Backport msg size limit 10.0.5

soujay:
This PR is a manual back port of the following commits that were made to master

- Add Azurite gitignore to src (#1016)
- Make sure messages close to the size limit can be moved as is to the error queue (#1002)
- Mention Azure Storage Explorer(#1014)

### DIFF
--- a/src/Transport/AtLeastOnceReceiveStrategy.cs
+++ b/src/Transport/AtLeastOnceReceiveStrategy.cs
@@ -6,9 +6,13 @@ namespace NServiceBus.Transport.AzureStorageQueues
     using System.Threading.Tasks;
     using Azure.Transports.WindowsAzureStorageQueues;
     using Extensibility;
+    using global::Azure;
     using Logging;
     using Transport;
 
+    /// <summary>
+    /// This corresponds to the RecieveOnly transport transaction mode
+    /// </summary>
     class AtLeastOnceReceiveStrategy : ReceiveStrategy
     {
         public AtLeastOnceReceiveStrategy(Func<MessageContext, Task> pipeline, Func<ErrorContext, Task<ErrorHandleResult>> errorPipe, CriticalError criticalError)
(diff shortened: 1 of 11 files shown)


src/Transport/AtLeastOnceReceiveStrategy.cs
```diff
@@ -55,6 +59,24 @@ public override async Task Receive(MessageRetrieved retrieved, MessageWrapper me
...
+                        // For an immediate retry, the error is logged and the message is returned to the queue to preserve the DequeueCount.
+                        // There is no in memory retry as scale-out scenarios would be handled improperly.
+                        Logger.Warn("Azure Storage Queue transport failed pushing a message through pipeline. The message will be requeued", ex);
+                        await retrieved.Nack().ConfigureAwait(false);
+                    }
+                    else
+                    {
+                        // Just acknowledge the message as it's handled by the core retry.
+                        await retrieved.Ack().ConfigureAwait(false);
+                    }
```
boblangley, line 73: I would move this back. And (see below) add a return to the new catch statement similar to the other exception. This will preserve the try/catch in the MessagePump.

src/Transport/AtLeastOnceReceiveStrategy.cs
```diff
@@ -55,6 +59,24 @@ public override async Task Receive(MessageRetrieved retrieved, MessageWrapper me
...
+                }
+                catch (RequestFailedException e) when (e.Status == 413 && e.ErrorCode == "RequestBodyTooLarge")
+                {
+                    Logger.WarnFormat($"Message with native ID `{message.Id}` could not be moved to the error queue with additional headers because it was too large. Moving to the error queue as is.", e);
+
+                    await retrieved.MoveToErrorQueueWithMinimalFaultHeaders(context, cancellationToken).ConfigureAwait(false);
```
boblangley, line 79:
```suggestion
                    await retrieved.MoveToErrorQueueWithMinimalFaultHeaders(context, cancellationToken).ConfigureAwait(false);
                    
                    return;
```

boblangley (changes requested):

boblangley (approved):

PR #1074, opened 2023-09-14
Address the clock drift in the assert visibility timeout

danielmarbach:
Fixes https://github.com/Particular/NServiceBus.AzureStorageQueues/issues/1075

This addresses the clock drift problem introduced in https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/138 by using the server time that is returned when retrieving messages and comparing it to the time that has passed based on the high precision stop watch at the time the `Ack` or `Nack` is called.

For the backports I will need to bring back the ValueStopwatch


src/Transport/MessageRetrieved.cs
```diff
@@ -102,14 +105,15 @@ public async Task MoveToErrorQueue(string messageId, string messagePopReceipt, B
...
-                var visibleIn = rawMessage.NextVisibleOn.Value - DateTimeOffset.UtcNow;
-                if (visibleIn < TimeSpan.Zero)
-                {
-                    var visibilityTimeoutExceededBy = -visibleIn;
-                    throw new LeaseTimeoutException(rawMessage, visibilityTimeoutExceededBy);
-                }
+                return;
+            }
+
+            var visibleIn = rawMessage.NextVisibleOn.Value - (messageReceivedTime + timeProvider.GetElapsedTime(startTimestamp));
```
SzymonPobiega, line 113: Nitpicking: could you extract variables here to make the calculation logic easier to follow? Something like
```suggestion
            var processingTime = timeProvider.GetElapsedTime(startTimestamp);
            var processedAtAccordingToServer = messageReceivedTime + processingTime;
            var visibleIn = rawMessage.NextVisibleOn.Value - processedAtAccordingToServer;
```

SzymonPobiega (approved):
Have one nitpicking comment

tmasternak (approved):

PR #1123, opened 2024-02-09
Remove use of newtonsoft json and prepare for RTM

andreasohlund:


andreasohlund:
@bording I’ve removed all newtonsoft stuff now

bording (approved):

PR #1212, opened 2024-09-06
Track messages that successfully completed the message or error pipeline but failed to get acknowledged due to expired leases

danielmarbach:
This PR follows through on related changes done in [Amazon SQS](https://github.com/Particular/NServiceBus.AmazonSQS/pull/1861) but adjusts the core pieces of the change to Azure Storage Queue-specific behavior. Additionally, it brings in a number of layered tests that verify the code works as expected and doesn't accidentally break later.

And LRU cache to track the delivery attempts is not necessary because the receive count is automatically updated by the storage queue infrastructure.

The logic in `AtLeastOnceReceiveStrategy` has been updated significantly:

- Caching Acknowledged Messages: A cache (using `BitFaster.Caching.Lru`) is introduced to keep track of messages that have been marked as successfully completed. This helps in avoiding unnecessary retries for messages that are already acknowledged.
- Lease Timeout Handling: The code now explicitly handles cases where a message's lease expires (via the existing `LeaseTimeoutException`). This is handled both when acknowledging a message (via `Ack()`) and during error recovery. If a lease timeout occurs, the message is tracked for completion in the next receive attempt.
- Error Handling Improvements: Enhanced handling around retries and error recovery, including better logging and recovery actions for different failure scenarios.
- Pipeline Changes: The handling pipeline now passes the raw Azure Queue messages (`QueueMessage`) through the context bag for further processing.

## Before

```mermaid
graph TD
    A[Receive Message] --> B[Invoke OnMessage]
    B --> C{Success?}
    C -- Yes --> D[Acknowledge message]
    D --> E[Message acknowledged]  
    C -- No --> F[Invoke OnError]
    F --> G{Error requires retry?}
    G -- Yes --> H[Requeue message]
    G -- No --> I[Acknowledge message]
    H --> J[Message requeued]
    I --> E[Message acknowledged]
```

## After

```mermaid
graph TD
    A[Receive Message] --> B[ID in LRU]
    B -- Yes --> M[Acknowledge message]
    F --> G[Message tracked]
    B -- No --> H[Invoke OnMessage]
    H --> I{Success?}
    I -- Yes --> M
    I -- No --> J[Invoke OnError]
    J --> K{Error requires retry?}
    K -- Yes --> L[Requeue message]
    K -- No --> M[Acknowledge message]
    M --> N{Ack success?}
    N -- Yes --> E[Message acknowledged]
    N -- No --> F[Track ID in LRU]
    G -- Message requeued --> A
```

## Tradeoffs

The tradeoffs are the sames as the other fixes applied in RabbitMQ and Amazon SQS. In competing consumer scenarios there can still be more retries because the LRU cache is in memory, but it is still marginally better than having a potentially unlimited number of retries. 

As for the transaction mode `None`, there is no change required since we are not giving any guarantees there and messages are acknowledged early on.

### DIFF
--- a/src/Transport/AtLeastOnceReceiveStrategy.cs
+++ b/src/Transport/AtLeastOnceReceiveStrategy.cs
@@ -30,7 +30,7 @@ namespace NServiceBus.Transport.AzureStorageQueues
             {
                 try
                 {
-                    Logger.DebugFormat("Received message (ID: '{0}') was marked as successfully completed. Trying to immediately acknowledge the message.", message.Id);
+                    Logger.DebugFormat("Received message (ID: '{0}') was marked as successfully completed. Trying to immediately acknowledge the message without invoking the pipeline.", message.Id);
                     await retrieved.Ack(cancellationToken).ConfigureAwait(false);
                 }
                 // Doing a more generous catch here to make sure we are not losing the ID and can mark it to be completed another time
--- a/src/Transport/AtLeastOnceReceiveStrategy.cs
+++ b/src/Transport/AtLeastOnceReceiveStrategy.cs
@@ -32,7 +32,7 @@ namespace NServiceBus.Transport.AzureStorageQueues
             }
 
             Logger.DebugFormat("Pushing received message (ID: '{0}') through pipeline.", message.Id);
-            var body = message.Body ?? Array.Empty<byte>();
+            var body = message.Body ?? [];
             var contextBag = new ContextBag();
             try
             {
--- a/src/Transport/AtLeastOnceReceiveStrategy.cs
+++ b/src/Transport/AtLeastOnceReceiveStrategy.cs
@@ -43,7 +43,7 @@ namespace NServiceBus.Transport.AzureStorageQueues
             }
             catch (LeaseTimeoutException)
             {
-                messagesToBeAcked.AddOrUpdate(message.Id, true);
+                TrackMessageToBeCompletedOnNextReceive();
             }
             catch (Exception ex) when (!ex.IsCausedBy(cancellationToken))
             {
@@ -69,7 +69,7 @@ namespace NServiceBus.Transport.AzureStorageQueues
                         }
                         catch (LeaseTimeoutException)
                         {
-                            messagesToBeAcked.AddOrUpdate(message.Id, true);
+                            TrackMessageToBeCompletedOnNextReceive();
                         }
                     }
                 }
@@ -93,6 +93,12 @@ namespace NServiceBus.Transport.AzureStorageQueues
                     }
                 }
             }
+
+            return;
+
+            void TrackMessageToBeCompletedOnNextReceive() =>
+                // The raw message ID might not be stable across retries, so we use the message wrapper ID instead.
+                messagesToBeAcked.AddOrUpdate(message.Id, true);
         }
 
         readonly OnMessage onMessage;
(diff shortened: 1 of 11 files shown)


src/Transport/AtMostOnceReceiveStrategy.cs
```diff
@@ -28,8 +29,10 @@ public override async Task Receive(MessageRetrieved retrieved, MessageWrapper me
         {
             Logger.DebugFormat("Pushing received message (ID: '{0}') through pipeline.", message.Id);
             await retrieved.Ack(cancellationToken).ConfigureAwait(false);
-            var body = message.Body ?? Array.Empty<byte>();
+            var body = message.Body ?? [];
             var contextBag = new ContextBag();
+            contextBag.Set<QueueMessage>(retrieved);
```
danielmarbach, line 34: This effectively crosses the boundaries of a feature. By adding it here it enables more advanced acceptance test scenarios, but it is also something customers can start leveraging, similar to https://docs.particular.net/transports/azure-service-bus/native-message-access. There are multiple ways how we can handle this. Happy to discuss further.

danielmarbach:
It might be worthwhile to stop passing the cancellation token in a few places before we approve this PR

danielmarbach:
> It might be worthwhile to stop passing the cancellation token in a few places before we approve this PR

we looked at the released versions of the transport and recognized the cancellation token is always passed to the ack or nack methods. Keeping that consistent for now. We will not address this but raise an improvement issue.

TravisNickels (approved):

src/Transport/AtLeastOnceReceiveStrategy.cs
```diff
@@ -36,9 +56,7 @@ public override async Task Receive(MessageRetrieved retrieved, MessageWrapper me
             }
             catch (LeaseTimeoutException)
             {
-                // The lease has expired and cannot be used any longer to Ack or Nack the message.
-                // see original issue: https://github.com/Azure/azure-storage-net/issues/285
-                throw;
+                TrackMessageToBeCompletedOnNextReceive();
```
soujay, line 59:
```suggestion
                TrackMessageToBeCompletedOnNextReceive();
```
```suggestion
                TrackMessageToBeCompletedOnNextReceive();
                
                throw;
```

soujay, line 59:
```suggestion
                TrackMessageToBeCompletedOnNextReceive();
                throw;
```

soujay: suggesting adding in a _throw_ so that it gets caught at https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/master/src/Transport/MessageReceiver.cs#L209

src/Transport/AtLeastOnceReceiveStrategy.cs
```diff
@@ -52,24 +70,31 @@ public override async Task Receive(MessageRetrieved retrieved, MessageWrapper me
...
-                        // Just acknowledge the message as it's handled by the core retry.
-                        await retrieved.Ack(cancellationToken).ConfigureAwait(false);
+                        try
+                        {
+                            // Just acknowledge the message as it's handled by the core retry.
+                            await retrieved.Ack(cancellationToken).ConfigureAwait(false);
+                        }
+                        catch (LeaseTimeoutException)
+                        {
+                            TrackMessageToBeCompletedOnNextReceive();
```
soujay, line 85:
```suggestion
                            TrackMessageToBeCompletedOnNextReceive();
                            throw;
```

soujay: suggesting adding in a throw so that it gets caught at https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/master/src/Transport/MessageReceiver.cs#L209

soujay (approved):

src/Transport/AtLeastOnceReceiveStrategy.cs
danielmarbach: I'm OK with that to get the log warn. This change though also requires some test changes

PR #1264, opened 2024-12-05
handle wrapped native message on a delayed retry

### LINKED ISSUE #1005: When using a custom unwrapper delayed retries or forwarding to the error queue can transform the message into a poison message
#### Description

In a situation where a customer [has specified a custom unwrapper](https://docs.particular.net/transports/azure-storage-queues/native-integration#custom-envelope-wrapper), unless the customer has implemented it carefully the delayed or faulted message will be wrapped in the default way by the transport which may be incompatible with the unwrapper and then it would become a poison message for the transport when a retry is attempted.

image

#### Expected behavior

The message can still be retried when delayed retries or retried from the error queue.

#### Actual behavior

The message(s) ends up a poison message hinder or even possibly clogging the queue.

#### Versions

Please list the version of the relevant packages or applications in which the bug exists.        

v9 and above.

### Steps to reproduce

1. Setup a native sender that wraps in a way incompatible with the default wrapping
1. Configure a custom wrapper without the ability to unwrap default wrapped messages
1. Send a native message
1. Ensure the message handler throws an exception

### Relevant log output

_No response_

#### Workarounds

Ensure the custom unwrapper can handle messages can be unwrapped that are wrapped by the default wrapper.

#### Possible solutions

Change the user wrapper internal implementation to automatically try the default unwrapper first.
or
Update the documentation to provide strong warnings to either always include the ability to unwrap default wrapped messages or turn off delayed retries/error queue forwarding.

PhilBastian:
fixes #1005

### DIFF
--- a/src/AcceptanceTests/DelayedDelivery/Native_message_delayed_retry.cs
+++ b/src/AcceptanceTests/DelayedDelivery/Native_message_delayed_retry.cs
@@ -1,6 +1,7 @@
 ﻿namespace NServiceBus.Transport.AzureStorageQueues.AcceptanceTests.DelayedDelivery
 {
     using System;
+    using System.Buffers.Text;
     using System.Collections.Generic;
     using System.Text.Json;
     using System.Threading;
@@ -12,7 +13,7 @@
     using NServiceBus.Azure.Transports.WindowsAzureStorageQueues;
     using NUnit.Framework;
 
-    class Nativedelayed : NServiceBusAcceptanceTest
+    class Native_message_delayed_retry : NServiceBusAcceptanceTest
     {
         [Test]
         public async Task Should_handle_delayed_delivery_of_native_message()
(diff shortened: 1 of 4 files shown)


src/AcceptanceTests/DelayedDelivery/Native_message_delayed_retry.cs
```diff
@@ -54,7 +55,9 @@ public SampleEndpoint() =>
                         var transport = cfg.ConfigureTransport<AzureStorageQueueTransport>();
                         transport.MessageUnwrapper = message =>
                         {
-                            return new MessageWrapper
+                            return Base64.IsValid(message.MessageText)
```
jpalac, line 58: Remind me what happens if the native message is a valid Base64 one? As it seems that this would then return null and we would use the default unwrapper which would fail?

PhilBastian: correct, but this is a test where the native message is only JSON serialised

abparticular (approved):

PR #1317, opened 2025-06-11
Fix race condition in RepeatedFailuresOverTimeCircuitBreaker

afprtclr:
Cherry-pick of https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/1315


.github/workflows/ci.yml
```diff
@@ -20,7 +20,7 @@ jobs:
         include:
           - os: windows-2019
             name: Windows
-          - os: ubuntu-20.04
+          - os: ubuntu-latest
```
WilliamBZA, line 23: Why are we changing the ubuntu version? I'm pretty sure this is the only downstream that would be using `ubuntu-latest` if we make this change

afprtclr: > Why are we changing the ubuntu version?

ubuntu-20.04 is not available and CI/CD doesn't work anymore.

> I'm pretty sure this is the only downstream that would be using ubuntu-latest if we make this change

That I don't know, but I had to do the same change in Service Control last month when I was releasing it.

WilliamBZA:
```suggestion
          - os: ubuntu-22.04
```

I think let's stick it to 22.04. [All other repos have that explicitly selected](https://github.com/search?q=org%3AParticular+ubuntu-22.04+language%3AYAML&type=code&l=YAML).

PR #1335, opened 2025-08-07
Remove custom PreObsolete attribute

tamararivera:
Follow-up to https://github.com/Particular/NServiceBus/pull/7358

This PR removes the repo copy of PreObsoleteAttribute and uses the one defined in Particular.Obsoletes instead.


src/Transport/MessageWrapper.cs
```diff
@@ -17,7 +18,8 @@ public class MessageWrapper : IMessage
         public string ReplyToAddress { get; set; }
 
         // Defining an obsoletion path for this has been raised on https://github.com/Particular/NServiceBus.AzureStorageQueues/issues/1318
-        [Obsolete("Legacy property for backwards compatibility.", error: false)]
+        [ObsoleteMetadata(TreatAsErrorFromVersion = "99", RemoveInVersion = "100")]
```
tamararivera, line 21: Workaround to set the versions.
I'm not attached to the numbers

PR #1461, opened 2026-06-11
Mark message-driven pub/sub compatibility mode as obsolete

mnadareski-particular:
All uses of `EnableMessageDrivenPubSubCompatibilityMode` need to be marked as obsolete with a warning starting in the next minor version.

This also adds the `#pragma warning disable/restore` directives for tests only.


mauroservienti (approved):
Cool 😎🚀

